PR #54, opened 2019-11-08
feat: remove tier from btp server query string

sharafian:
The server isn't doing anything with it so should be removed

### DIFF
--- a/packages/coil-extension/src/background/services/Stream.ts
+++ b/packages/coil-extension/src/background/services/Stream.ts
@@ -180,13 +180,8 @@ export class Stream extends EventEmitter {
     // is severed before full establishment
     if (!this._active) throw new Error('aborted monetization')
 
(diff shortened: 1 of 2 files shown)


packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -180,13 +180,8 @@ export class Stream extends EventEmitter {
...
     if (!this._active) throw new Error('aborted monetization')
 
-    const tier = await this._tiers.getTier(this._initiatingUrl)
-    const serverWithTier = `${this._server}?tier=${tier}`
-
-    if (!this._active) throw new Error('aborted monetization')
-
     this._plugin = new IlpPluginBtp({
-      server: serverWithTier,
+      server: this._server,
```
sublimator: Do you want me to kill the BandwidthTiers class entirely. It's no longer going to be used?

PR #90, opened 2019-11-19
fix(coil-extension): Stream race condition

sentientwaffle:
Previously, rapidly switching between monetized tabs could trigger an assortment of errors (e.g. `TypeError: Cannot read property 'isOpen' of undefined`, `Unhandled error. (undefined)`, some timeouts, `Error: Unexpected error while sending packet. Code: F00, triggered by: [address], message: Cannot destructure property `userId` of 'undefined' or 'null'.`), and weird states (e.g. multiple simultaneous streams).

Preventing multiple streams from running simultaneously fixes the errors.

### DIFF
--- a/packages/coil-extension/src/background/services/Stream.ts
+++ b/packages/coil-extension/src/background/services/Stream.ts
@@ -65,10 +65,8 @@ export class Stream extends EventEmitter {
   private _lastOutgoingMs!: number
   private _packetNumber!: number
   private _active: boolean
-  private _plugin!: IlpPluginBtp
-  private _ilpStream!: DataAndMoneyStream
-  private _connection!: Connection
-  private _bandwidth: AdaptiveBandwidth
+  private _looping: boolean
+  private _attempt: StreamAttempt | null
   private _client: GraphQlClient
   private _coilDomain: string
 
@@ -105,13 +103,10 @@ export class Stream extends EventEmitter {
     )
 
     this._active = false
+    this._looping = false
+    this._attempt = null
     this._lastDelivered = 0
     this._initiatingUrl = initiatingUrl
-    this._bandwidth = new AdaptiveBandwidth(
-      this._initiatingUrl,
-      this._tiers,
-      this._debug
-    )
 
     const server = new URL(this._coilDomain)
     server.pathname = '/btp'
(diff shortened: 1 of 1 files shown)


packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -211,14 +222,86 @@ export class Stream extends EventEmitter {
...
+  requestId: string
+}
+
+let ATTEMPT = 0
 
-    const [plugin, spspDetails] = await Promise.all([
-      this._getPlugin(),
-      this._getSPSPDetails()
-    ])
+class StreamAttempt {
```
sublimator: Nice. It's always kind of bugged me that the Stream class represented multiple streams and the way it handled state. Will look at this closer today.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -129,38 +124,55 @@ export class Stream extends EventEmitter {
...
-        } catch (e) {
-          this._debug('error streaming. retry in 2s. err=', e.message)
-          await this._cleanUpFailedStream()
-          await timeout(2000)
         }
+      } catch (e) {
+        this._debug('error streaming. retry in 2s. err=', e.message, e.stack)
+        if (this._active) await timeout(2000)
+      } finally {
+        if (attempt) bandwidth.addSentAmount(attempt.getTotalSent())
```
sublimator: attempt seems like it would always be defined here

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -129,38 +124,55 @@ export class Stream extends EventEmitter {
...
-      // reset our timer when we start streaming.
-      this._bandwidth.reset()
-
-      while (this._active) {
-        try {
-          await this._stream()
+    if (this._active) return
+    this._active = true
+    if (this._looping) return
+    this._looping = true
```
sublimator: I think I understand this. You could stop then restart so fast that multiple loops are in action at once right ?

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -211,14 +222,86 @@ export class Stream extends EventEmitter {
...
-    const [plugin, spspDetails] = await Promise.all([
-      this._getPlugin(),
-      this._getSPSPDetails()
-    ])
+class StreamAttempt {
+  private readonly _onMoney: (event: any) => void
+  private readonly _bandwidth: AdaptiveBandwidth
+  private readonly _debug: (...args: any[]) => void
+
+  private _plugin!: IlpPluginBtp
```
sublimator: These non null assertion operators were always the kind of smell that this wasn't factored properly. I mean even here, you could construct the object only once you have the _plugin and details ready.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -327,68 +404,60 @@ export class Stream extends EventEmitter {
     })
   }
 
-  onMoney(sentAmount: string) {
+  async stop(): Promise<void> {
+    this._active = false
+    if (!this._connection) return
+
+    this._debug('initiating stream shutdown')
+    if (this._ilpStream.isOpen()) {
```
sublimator: I think at some point (not necessarily this pull request) we should have this factored so that there's no doubt that there's no interleaving of operations. 

Was hoping to pull out a shared set of classes can use in oauth-scripts and the extension.

packages/coil-extension/src/background/services/Stream.ts
sentientwaffle: Yes, that's what I'm trying to avoid.

PR #115, opened 2019-11-27
perf: speed up monetizationstart

sentientwaffle:
In ilp-protocol-stream, `Stream`s emit `"outgoing_money"` events before `connection.totalDelivered` is updated. This resulted in the extension ignoring the first `outgoing_money` event, since it mistakenly assumed that no money was delivered.

In my tests, this shortened the time-to-`monetizationstart` from 2500-2800ms down to 500-700ms.

### DIFF
--- a/packages/coil-extension/src/background/services/Stream.ts
+++ b/packages/coil-extension/src/background/services/Stream.ts
@@ -331,7 +331,8 @@ class StreamAttempt {
 
     return new Promise((resolve, reject) => {
       const onMoney = (sentAmount: string) => {
-        this.onMoney(sentAmount)
+        // Wait until `nextTick` so that `connection.totalDelivered` has been updated.
+        process.nextTick(this.onMoney.bind(this), sentAmount)
       }
 
       const onPluginDisconnect = async () => {
--- a/packages/coil-extension/src/background/services/Stream.ts
+++ b/packages/coil-extension/src/background/services/Stream.ts
@@ -331,8 +331,8 @@ class StreamAttempt {
 
     return new Promise((resolve, reject) => {
       const onMoney = (sentAmount: string) => {
-        // Wait until `nextTick` so that `connection.totalDelivered` has been updated.
-        process.nextTick(this.onMoney.bind(this), sentAmount)
+        // Wait until `setImmediate` so that `connection.totalDelivered` has been updated.
+        setImmediate(this.onMoney.bind(this), sentAmount)
       }
 
       const onPluginDisconnect = async () => {
--- a/packages/coil-oauth-scripts/src/Stream.ts
+++ b/packages/coil-oauth-scripts/src/Stream.ts
@@ -173,7 +173,9 @@ export class Stream {
       stream.setSendMax(initialSendMaxAmount)
 
       await new Promise((resolve, reject) => {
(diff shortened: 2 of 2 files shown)


sharafian:
Is this speedup also after applying optimizations like exchange rate and packet size hints or have those not gone in yet? I would be curious if they can get the time down even further

sentientwaffle:
It's without, I think. I tested enabling them and they don't make as much of a different as I expected, so I'm actually looking into that next.

sharafian:
Is this fix applicable to the oauth script also? right now the two have different modules that send on streams

sentientwaffle:
I didn't think of that. Yes, I think it is applicable, I'll update the PR.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -331,7 +331,8 @@ class StreamAttempt {
 
     return new Promise((resolve, reject) => {
       const onMoney = (sentAmount: string) => {
-        this.onMoney(sentAmount)
+        // Wait until `setImmediate` so that `connection.totalDelivered` has been updated.
```
sublimator: Nice work Mr @sentientwaffle !

sublimator: Maybe ilp-protocol-stream should be updated to set that beforehand ?

sublimator:
@sentientwaffle 
I can't actually replicate your TTS reductions here in Cambodia. 
Were you testing against a local server at all ?

sublimator:
@sentientwaffle 
I replicated against a local ilp-plugin-mini-accounts server

packages/coil-extension/src/background/services/Stream.ts
sentientwaffle: Since updating the hold [may not change the amount](https://github.com/interledgerjs/ilp-protocol-stream/blob/762d4b36583ab70848b0f6d7fe69a4ac3d884644/src/stream.ts#L426) that isn't desirable.

sentientwaffle:
@sublimator Testing against a local server may not work because the packets would arrive so quickly that the delay may be too small to notice. I was testing against coil's staging server.

sublimator:
actually, the delay was noticeable on local server if that means anything
to you!?

will PM you to discuss soon

packages/coil-extension/src/background/services/Stream.ts
sublimator: Can you elaborate on that? 
Note that I'm not a STREAM/ILP expert.

sentientwaffle: You're right, stream could set the amount beforehand, but I think it would qualify as a breaking change (albeit just barely). I've made a note-to-self to include that change the next time I make a breaking stream change. I don't think it warrants one on its own since it's so simple to work around.

sublimator: Ok, so outgoing_money is when money is actually _delivered_ right? Or am I wrong there? 
It's definitely a change but one could reasonably argue it's actually a bug and the current behavior is incorrect. Would it really actually "break" any client code or simply give more correct results ?

sentientwaffle: Yes, `outgoing_money` is the notification that money has been successfully delivered on a stream.

PR #146, opened 2019-12-05
fix(oauth-scripts): quick fix

sublimator:


packages/coil-oauth-scripts/src/Stream.ts
```diff
@@ -71,7 +71,8 @@ export class Stream {
     this.monetization = new DocumentMonetization(document)
   }
 
-  start(opts: StreamStartOpts) {
+  async start(opts: StreamStartOpts) {
+    await this.finalizeExisting()
```
sublimator: @sharafian 
Wait for the last promise in the loop to finish before starting a new one. 
In practice we shouldn't be stopping/starting fast enough for there to be much need for a queue, though in theory there is an emergent one.   
In any case we will fix this properly with a new package.

PR #147, opened 2019-12-05
feat: query for payment pointer with channel URL

njlie:
Extension side of [coilhq/coil #2939](https://github.com/coilhq/coil/issues/2939).

### DIFF
--- a/packages/coil-extension/src/content/services/DomService.ts
+++ b/packages/coil-extension/src/content/services/DomService.ts
@@ -1,0 +1,42 @@
+import { timeout } from '../util/timeout'
+
+interface PollForYouTubeChannelIdParams {
+  selector: string
+  everyMs: number
+  times: number
+}
+
+export class DomService {
+  constructor(private document: Document, private window: Window) {}
+
+  async pollForElement<T extends HTMLElement = HTMLElement>({
+    selector,
+    everyMs,
+    times
+  }: PollForYouTubeChannelIdParams): Promise<T | null> {
+    let polls = 0
+    while (++polls <= times) {
+      const el = this.document.querySelector<T>(selector)
+      if (el) {
+        return el
+      } else {
+        await timeout(everyMs)
+      }
+    }
+    return null
+  }
+
+  async windowLoaded() {
+    if (this.document.readyState !== 'complete') {
+      await new Promise(resolve => {
+        this.window.addEventListener(
+          'load',
+          () => {
+            resolve()
+          },
+          { once: true }
+        )
+      })
+    }
+  }
+}
--- a/packages/coil-extension/src/content/services/DomService.ts
+++ b/packages/coil-extension/src/content/services/DomService.ts
@@ -11,7 +11,7 @@ export interface PollForElementParams {
 
 @injectable()
 export class DomService {
-  constructor(private document: Document, private window: Window) {}
+  constructor(private document: Document) {}
 
   async pollForElement<T extends HTMLElement = HTMLElement>({
     selector,
--- a/packages/coil-extension/src/content/services/DomService.ts
+++ b/packages/coil-extension/src/content/services/DomService.ts
@@ -1,4 +1,7 @@
+import { injectable } from 'inversify'
+
 import { timeout } from '../util/timeout'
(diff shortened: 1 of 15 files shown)


sublimator:
>  coil-puppeteer-utils:test test page failed to open stream. page= twitchUrl +1ms

Seems like the twitch test failed ?

packages/coil-extension/src/content/services/RunContentHandler.ts
```diff
@@ -16,6 +16,12 @@ interface GetPageData {
   }
 }
 
+interface HyperlinkElement extends Element {
```
sublimator: There's HTMLAnchorElement

packages/coil-extension/src/content/services/RunContentHandler.ts
```diff
@@ -28,18 +34,43 @@ export class RunContentHandler {
     private client: GraphQlClient
   ) {}
 
-  async adaptedPageDetails(url: string) {
+  async adaptedPageDetails(url: string, site: string) {
     debug('fetching payment pointer for this page', url)
 
+    // Wait for page to load to ensure there's a channel url to grab
+    if (document.readyState !== 'complete' && site === 'youtube') {
```
sublimator: Try and use document/window via DI so as this gets restructured it's in a good shape for tests with mocking

packages/coil-extension/src/content/services/RunContentHandler.ts
```diff
@@ -16,6 +16,12 @@ interface GetPageData {
   }
 }
 
+interface HyperlinkElement extends Element {
+  href: string
+}
+
+const CHANNEL_ID_REGEX = /https?:\/\/(www\.)?youtube\.com\/channel\//i
```
sublimator: This appears unused ?

packages/coil-extension/src/content/services/RunContentHandler.ts
```diff
@@ -24,22 +29,40 @@ export class RunContentHandler {
...
+    debug('fetching payment pointer for this page', url, site)
+
+    const variables = { url }
+
+    if (site === 'youtube') {
+      // Wait for page to load and the element containing the channelId is
+      // available
+
+      await this.windowLoaded()
+      const channelUrl = await this.pollForYouTubeChannelId({
```
sublimator: @njlie 
I pushed a commit while I was testing this out. Seems sometimes the element isn't available until a bit after load.

packages/coil-extension/src/content/services/RunContentHandler.ts
```diff
@@ -54,6 +77,47 @@ export class RunContentHandler {
...
+
+  async windowLoaded() {
+    if (this.document.readyState !== 'complete') {
+      await new Promise(resolve => {
+        this.window.addEventListener(
+          'load',
+          () => {
+            resolve()
+          },
+          { once: true }
```
sublimator: @njlie 
Probably doesn't *really* matter but could potentially add a bunch of persistent listeners given youtube uses HTML5 pushState so can use once:true

packages/coil-extension/src/content/services/RunContentHandler.ts
njlie: Yeah, was in an earlier attempt that I forgot to remove.

njlie:
> > coil-puppeteer-utils:test test page failed to open stream. page= twitchUrl +1ms
> 
> Seems like the twitch test failed ?
> 
> edit: is that because it requires backend features that aren't deployed to production yet ? the new graphql query?

Yeah, we're gonna need to roll out some API changes first before this will work.

Also, thanks for updating the branch.

packages/coil-extension/src/content/services/DomService.ts
```diff
@@ -0,0 +1,42 @@
...
+        await timeout(everyMs)
+      }
+    }
+    return null
+  }
+
+  async windowLoaded() {
+    if (this.document.readyState !== 'complete') {
+      await new Promise(resolve => {
+        this.window.addEventListener(
```
sublimator: wait, what about document listener for DOMContentLoaded ?

PR #187, opened 2019-12-12
perf: slippage=1, fixed exchangeRate,maximumPacketAmount

sentientwaffle:
May as well hold off on releasing this until it includes https://github.com/interledgerjs/ilp-protocol-stream/pull/124.

### DIFF
--- a/packages/coil-extension/src/background/services/Stream.ts
+++ b/packages/coil-extension/src/background/services/Stream.ts
@@ -323,7 +323,9 @@ class StreamAttempt {
     this._connection = await createConnection({
       ...this._spspDetails,
       plugin,
-      slippage: 0.05
+      slippage: 1.0,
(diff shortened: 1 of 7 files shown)


packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -323,7 +323,9 @@ class StreamAttempt {
     this._connection = await createConnection({
       ...this._spspDetails,
       plugin,
-      slippage: 0.05
+      slippage: 1.0,
```
sublimator: We will want to do this for the oauth-scripts too I assume

sentientwaffle: @sublimator updated. What's the best way to test oauth-scripts locally (pref. against staging)?

PR #221, opened 2019-12-19
Authenticate with anonymous tokens

sharafian:
Still an incomplete draft. This feature builds a store of anonymous tokens generated through a blind signature scheme. The tokens will be collected from Coil's API while logged in, and then submitted anonymously behind a proxy to authenticate Interledger traffic.

Most of the tricky logic is around managing these tokens. Each token is worth a standard and small amount of money, in order to prevent fingerprinting. But this also means we need to be able to swap tokens while on a single page and manage which have been consumed while keeping efficiency. We want to request enough tokens up front to throw off timing attacks, but not request so many that clearing your localstorage wipes out your usage for the month.

-----------
- [ ] ideally, the code in the coil-extension part should be behind a feature flag or something so that this work can get merged before we're ready to roll it out. Otherwise we'll likely have to go through and refresh it with latest code changes when we actually want to use it.

- [ ] At first startup, the extension should request a batch of tokens from the redeemer service. This should be worth about an hour of Coil usage. Each token could be usable for around a minute of Coil usage. The redeemer service defines the value of a token which can be referred to for this

- [ ] Whenever a WM page is visited, the Streams service should grab an anonymous token + signature from its batch of anonymous tokens instead of fetching it via the normal Coil API. It submits the anonymous token + signature against the redeemer service to redeem it for a JWT. The JWT can be used to authenticate a BTP connection in the same way as a normal BTP token.

- [ ] If the anonymous token is exhausted, the redeemer service will respond with an insufficient liquidity error with the message "exhausted capacity." https://github.com/coilhq/coil/pull/2802/files#diff-c31e4e873609785bd81a341808dafffbR333 The token which has been exhausted should be removed from the token store and a new anonymous token should be rotated in (redeemed for JWT, etc). This will require restarting the STREAM connection.

- [ ] When the number of anonymous tokens is reduced below some threshold (10% could be a starting threshold), a new batch of anonymous tokens should be requested. The size of this batch requested should be the same as the initial batch to leak less information.

- [ ] If the redeemer service rejects a request for an anonymous token due to exhausted usage, then we need to handle that case.
   - This could work by just ceasing monetization and showing a warning state. I don't think we'd want to roll that out, but it could at least let this feature get developed until we work out a better solution.
   - It could change into a state where monetization events are emitted with no actual micropayments occurring (and no monetizationprogress events emitted either). Basically a way of saying "I'm monetized but not paying right now," similar to the low throughput mode currently experienced when you use Coil too much. This lays the groundwork for a future state where micropayments can be accounted by provider instead of by individual session ID. 
https://github.com/adrianhopebailie/open-payments/issues/19 In this future state a user who exhausted their usage could just show the site an attestation that they belong to their provider and access content off of the excess paid by other members of that provider. Not sure I would want to roll out this version either but it'd be a lot closer to the future state where we expect to roll this out.
  - Another option is to let the redeemer service offer two tiers of usage, one which represents ~1 min of usage at normal bandwidth and one that represents ~1min of usage at reduced bandwidth. It would be highly unlikely that we would run out of reduced bandwidth tokens


packages/coil-anonymous-tokens/src/index.ts
```diff
@@ -0,0 +1,151 @@
...
+    fn: (
+      value: string,
+      key: string,
+      iterationNumber: number
+    ) => SignedToken | undefined
+  ) => Promise<SignedToken | undefined>
+}
+
+export class AnonymousTokens {
+  private store: TokenStore
```
adrianhopebailie: This can use constructor assignment

```suggestion
  constructor(private store: TokenStore) {
  }
```

sharafian:
Update: got this to work issuing/redeeming tokens against our redeemer service that sits in front of https://github.com/privacypass/challenge-bypass-server . The code for issuing/redeeming on the client side is based off of https://github.com/privacypass/challenge-bypass-extension

Still todo:

- [ ] clean up imported code (maybe break it out into a standalone module that can also be used for tests on the server side)
- [ ] move some of the privacy-pass specific functions in index down into lib/privacypass
- [ ] don't hard-code config options for privacy pass library
- [ ] right now the commitment from the server is hard-coded off of the test commitment. We should code in real commitments or fetch them from our redeemer service dynamically
- [ ] fix the tests so that they either don't run in CI or are able to run against a redeemer service instance
- [ ] more comprehensive tests

PR #257, opened 2020-01-10
feat(coil-extension): detect system clock skew

sublimator:


packages/coil-extension/src/popup/components/MonetizedPage.tsx
```diff
@@ -72,24 +72,49 @@ export function MonetizedPage(props: PopupProps) {
 }
 
 function Donating(props: PopupProps) {
-  const { monetizedTotal } = props.context.store
+  const { monetizedTotal, error } = props.context.store
+  const onTimeIsClicked = useCallback(
```
sublimator: seems pointless

traviscrist:
@sublimator Can we revert this out of master? I want to run this by Ben when he gets back since I ideally think we can come up with a better solution to the clock skew issue than just showing users the time.is and I'd prefer to keep this out of master until some decision is made on the original github issue #2903

PR #269, opened 2020-01-12
feat(coil-extension): update popup copy

sublimator:
Extension side of https://github.com/coilhq/coil/pull/2821

### DIFF
--- a/packages/coil-extension/src/popup/components/Unsubscribed.tsx
+++ b/packages/coil-extension/src/popup/components/Unsubscribed.tsx
@@ -7,10 +7,10 @@ import { PopupProps } from '../types'
 import { StatusButton } from './StatusButton'
 import { StatusTypography } from './util/StatusTypography'
 
-const titleString = 'Get a subscription'
-const subheading1 = 'To use Coil you need an active subscription'
+const titleString = 'Get a membership'
+const subheading1 = 'To use Coil you need an active membership'
 const footerString =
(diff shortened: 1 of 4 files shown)


packages/coil-extension/src/popup/components/Unsubscribed.tsx
```diff
@@ -7,10 +7,10 @@ import { PopupProps } from '../types'
 import { StatusButton } from './StatusButton'
 import { StatusTypography } from './util/StatusTypography'
 
-const titleString = 'Get a subscription'
-const subheading1 = 'To use Coil you need an active subscription'
+const titleString = 'Get a membership'
```
sublimator: This seems kind of awkward. Is the plan to use membership on the site as well?

sublimator: image

^ This will become Membership too ?

traviscrist: The site will change over to membership once the new settings page Dacio is working on is merged in.

traviscrist:
@fruehle does `Get a membership` sound ok to you, its for the extension state where you are signed into coil but don't have an active membership.

fruehle:
@travis: We used ‘Become a member’ on the static pages for the CTA.

sublimator:
sounds less awkward

traviscrist:
@fruehle that sounds much better, @sublimator lets go with `Become a member`

PR #284, opened 2020-01-16
fix: explicitly @inject default param, fixes #206

### LINKED ISSUE #206: Extension not working against staging (when using TRANSPILE_ONLY)
- [x] Resolve issues with staging (likely cloudflare access related)

sublimator:


packages/coil-client/src/graphQlClient.ts
```diff
@@ -29,6 +29,7 @@ export class GraphQlClient {
   public whoAmI = whoAmI
 
   public constructor(
+    @inject(GraphQlClientOptions)
```
sublimator: This also fixes:
```
  public constructor(
    // @inject(GraphQlClientOptions)
    private config: GraphQlClientOptions//  = new GraphQlClientOptions()
  ) {
```
It's something to do with the default value

PR #286, opened 2020-01-16
misc anonymous token updates

sentientwaffle:
Note that this is a PR to `bs-anonymous-tokens`, not `master`. Some of these changes depend on https://github.com/coilhq/coil/pull/3071

Open questions:

* Right now to determine if a token is exhausted the extension checks for the substring `"exhausted capacity"` in the error message. Is that Good Enough or should a specific error code be selected (right now it uses `F00`)?
* What domains will the redeemer and signer services run on?

### DIFF
--- a/packages/coil-extension/src/background/services/Stream.ts
+++ b/packages/coil-extension/src/background/services/Stream.ts
@@ -188,7 +188,13 @@ export class Stream extends EventEmitter {
           await timeout(1000)
         }
       } catch (e) {
-        if (e.message.includes('exhausted capacity.') && btpToken) {
+        const { ilpReject } = e
+        if (
+          btpToken &&
+          ilpReject &&
+          ilpReject.message === 'exhausted capacity.' &&
+          ilpReject.data.toString() === btpToken
+        ) {
           this._debug('anonymous token exhausted; retrying, err=%s', e.message)
           this._anonTokens.removeToken(btpToken)
           continue
--- a/packages/coil-extension/src/background/services/Stream.ts
+++ b/packages/coil-extension/src/background/services/Stream.ts
@@ -27,7 +27,6 @@ import { Logger, logger } from './utils'
 
 const { timeout } = asyncUtils
 
-const ANON_TOKEN_BATCH_SIZE = 10
 const UPDATE_AMOUNT_TIMEOUT = 2000
 let ATTEMPT = 0
 
@@ -170,7 +169,7 @@ export class Stream extends EventEmitter {
       let btpToken: string | undefined
       let plugin, attempt
       try {
-        btpToken = await this._getBtpToken()
+        btpToken = await this._anonTokens.getToken(this._authToken)
         plugin = await this._makePlugin(btpToken)
         const spspDetails = await this._getSPSPDetails()
         this.container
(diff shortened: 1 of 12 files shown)


sublimator:
@sentientwaffle 
> I don't understand inversify very well, but since it wouldn't be able to use @inject what would even be the point?

Well it depends what other clients there are for the code. It's not really a big deal, but you can use the lightweight annotations without having to drag in all of the IoC. Basically the idea behind https://docs.oracle.com/javaee/6/api/javax/inject/package-summary.html

https://github.com/coilhq/web-monetization-projects/blob/master/packages/coil-client/src/graphQlClient.ts#L1-L32

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -197,19 +206,20 @@ export class Stream extends EventEmitter {
...
-  async _getBtpToken() {
-    try {
-      this._debug('fetching btp token')
-      return await this._client.refreshBtpToken(this._token)
-    } catch (e) {
-      this._debug('error fetching btp token', e.message)
-      throw e
-    }
-  }
+  async _getBtpToken(): Promise<string> {
```
sublimator: Not sure when you guys wanted to roll all this out, but could potentially develop on master behind a feature flag (WEBPACK_DEFINE_XXX) if ya felt it was more convenient in any way.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -185,6 +189,11 @@ export class Stream extends EventEmitter {
           await timeout(1000)
         }
       } catch (e) {
+        if (e.message.includes('exhausted capacity.') && btpToken) {
```
sharafian: is there anything we can do to verify that this is actually coming from the redeemer service? we could start by looking at ILP address or maybe even have it a signed message in the future

My worry is just that if I set up a malicious SPSP server that returns `F00 exhausted capacity` on every packet, then people who visited my site would burn through all of their WM tokens

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -197,19 +206,20 @@ export class Stream extends EventEmitter {
...
-      throw e
-    }
-  }
+  async _getBtpToken(): Promise<string> {
+    const cachedAnonToken = await this._anonTokens.getToken()
+    if (cachedAnonToken) return cachedAnonToken
 
-  async _makePlugin() {
-    const btpToken = await this._getBtpToken()
+    await this._anonTokens.populateTokens(
```
sharafian: we might want to populate tokens early, like somehow returning a warning from `getToken` if you're running low. Then we don't need to put a costly populateTokens call on the critical path to web monetizing a page

We should also probably have some kind of lock in the token service to prevent two populate calls from running at once. If I switch between WM tabs while populateTokens is running then it might start another populateTokens call and get me too many tokens

packages/coil-extension/src/background/services/Stream.ts
sentientwaffle: Isn't a rejection's `triggeredAt` is just as easily spoofed as the `message`?

Returning a signed message in the data would work. Alternatively, the error could be a BTP-level error instead of an ILP-level one (i.e. using a BTP error packet). I'm not sure if that would make sense from a protocol point of view, but rogue SPSP servers couldn't spoof it.

sharafian: perhaps the signature is the best way to go, then. Or because there is not hop between the extension and Coil, maybe it could even return the btpToken to prove that it's really Coil.

sentientwaffle:
Depends on https://github.com/coilhq/coil/pull/3099

Issue #144, opened 2019-12-05
Extension time syncing issue


sublimator:
https://github.com/coilhq/coil/issues/2903

sublimator:
@sentientwaffle 
You got this one right ? STREAM intricacies over my head

sentientwaffle:
Yes

sentientwaffle:
Resolved by https://github.com/coilhq/web-monetization-projects/pull/307.

PR #307, opened 2020-01-22
fix(coil-extension): fix clock skew

sentientwaffle:
Ensure that payment succeeds in spite of a client's clock skew.

Send all Prepare packets with a fixed `expiresAt` in the far future (`2100-01-01T00:00:00.000Z`). The connector clamps that using the `maxHoldTime` (30s) before forwarding the Prepare.

Note that this includes https://github.com/interledgerjs/ilp-protocol-stream/pull/128 and https://github.com/interledgerjs/interledgerjs/pull/7.

Fixes https://github.com/coilhq/coil/issues/2903.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2496,7 +2496,7 @@
   dependencies:
     "@types/node" "*"
 
-"@types/long@^4.0.0":
+"@types/long@4.0.0", "@types/long@^4.0.0":
   version "4.0.0"
   resolved "https://registry.yarnpkg.com/@types/long/-/long-4.0.0.tgz#719551d2352d301ac8b81db732acb6bdc28dbdef"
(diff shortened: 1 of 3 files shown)


sublimator:
@sentientwaffle 
Is this good to go in terms of the backend ?

yarn.lock
```diff
@@ -8403,6 +8403,14 @@ ilp-packet@^3.0.8, ilp-packet@^3.0.9:
     extensible-error "^1.0.2"
     oer-utils "^5.0.0"
 
+ilp-packet@^3.1.0-alpha.0:
```
sublimator: monorepo releases defaulting to alpha huh

sublimator:
related issue: https://github.com/coilhq/web-monetization-projects/issues/144

yarn.lock
sentientwaffle: Yeah, I don't quite have the hang of lerna monorepo management. This version works fine, though.

sentientwaffle:
@sublimator yes, connectors will handle this correctly. They just need an `"expiry"` middleware, which is enabled by default.

PR #310, opened 2020-01-23
feat(coil-puppeteer-utils): add CF-Access-* headers

sublimator:


packages/coil-puppeteer-utils/src/lib/initCoil.ts
```diff
@@ -32,6 +33,8 @@ export async function initCoil({
   password
 }: InitCoilParameters): Promise<InitCoilReturn> {
   const page = await browser.newPage()
+  await addCloudFlareAccessHeaders(page)
```
sublimator: Only really need this here once, thereafter the CF_Authorization cookie is set

PR #438, opened 2020-03-04
feat: tipping preview

sharafian:
For users enabled for tipping, this adds a button to the popup on Web Monetized pages. Right now it's an internal pilot that sends out one dollar at a time. This does not affect the experience for users on which `canTip` is not true (i.e. they aren't `@coil.com`)

### DIFF
--- a/packages/coil-extension/src/background/services/AuthService.ts
+++ b/packages/coil-extension/src/background/services/AuthService.ts
@@ -65,7 +65,6 @@ export class AuthService extends EventEmitter {
     if (resp.data?.whoami) {
       this.store.user = {
         ...resp.data.whoami,
-        canTip: true
       }
       return token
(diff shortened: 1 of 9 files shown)


packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -63,7 +63,9 @@ export class AuthService extends EventEmitter {
     const resp = await this.client.whoAmI(token)
     this.log('updateWhoAmi resp', resp.data)
     if (resp.data?.whoami) {
-      this.store.user = resp.data.whoami
+      this.store.user = {
+        ...resp.data.whoami
```
sublimator: Does this do anything? It's not a deep copy, so you still referencing the gizzards of the object

PR #520, opened 2020-04-09
feat: dont unnecessarily query whoami

sentientwaffle:
Don't unnecessarily query the user/subscription -- wait until the subscription/trial expires. This is more efficient, and prevents correlation with token redeeming.

### DIFF
--- a/packages/coil-extension/src/background/services/AuthService.ts
+++ b/packages/coil-extension/src/background/services/AuthService.ts
@@ -39,20 +39,21 @@ export class AuthService extends EventEmitter {
       token = null
     } else if (tokenUtils.isExpired({ token, withinHrs: 12 })) {
       // Update the stored token/user
-      this.log('doing refreshTokenAndUpdateWhoAmi')
+      this.log('before refreshTokenAndUpdateWhoAmi')
       token = await this.refreshTokenAndUpdateWhoAmi(token)
-      this.log('token after refreshTokenAndUpdateWhoAmi', token)
+      this.log('after refreshTokenAndUpdateWhoAmi', token)
     } else {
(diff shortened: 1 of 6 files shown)


packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -39,20 +39,21 @@ export class AuthService extends EventEmitter {
...
-      token = await this.cached.query({
-        key: 'updateWhoAmI',
-        maxAgeMs: 15e3,
-        operationFactory: async () => this.updateWhoAmi(stored)
-      })
-      this.log('token after updateWhoAmI', token)
+      const endDate =
+        this.store.user?.subscription?.endDate ||
+        this.store.user?.subscription?.trialEndDate
+      if (!endDate || new Date(endDate) < new Date()) {
```
sublimator, line 53: I think the reason why this was doing the query every time was because of the case where you had no subscription, then subscribed. It seems like this should handle that case. The `!endDate` branch.

sentientwaffle: No, before this change it was re-fetching the whoami any time it reconnected (unless it was within 15 seconds of the last fetch).

sublimator: > re-fetching the whoami any time it reconnected
> I think the reason why this was doing the query every time was because of the case where you had no subscription, then subscribed. 

Yeah. Sorry, I wasn't very clear today. I think we are saying the same things.

sublimator: The testing checklist: 
- [ ] Check can stream immediately after subscribing

PR #567, opened 2020-05-04
Add CustomEvent for Tips

sabinebertram:
On click of "Tip this creator $1!" within the extension, a CustomEvent called 'tip' is emitted, in the same fashion as 'montizationstart' or 'monetizationprogress'. It shows an estimated amount in the receiver's assetCode and assetScale.

### DIFF
--- a/packages/coil-extension/src/background/services/BackgroundScript.ts
+++ b/packages/coil-extension/src/background/services/BackgroundScript.ts
@@ -23,7 +23,8 @@ import {
   SetMonetizationState,
   SetStreamControls,
   StartWebMonetization,
-  ToBackgroundMessage
+  ToBackgroundMessage,
+  Tip
 } from '../../types/commands'
 import { LocalStorageProxy } from '../../types/storage'
 import { TabState } from '../../types/TabState'
@@ -674,6 +675,13 @@ export class BackgroundScript {
         }
       })
       this.log(`sendTip: sent tip to ${receiver}`, result)
+      const message: Tip = {
+        command: 'tip',
+        data: {
+          paymentPointer: receiver
+        }
+      }
+      this.api.tabs.sendMessage(tabId, message)
       return { success: true }
     } catch (e) {
       this.log(`sendTip: error. msg=${e.message}`)
--- a/packages/coil-extension/src/background/services/BackgroundScript.ts
+++ b/packages/coil-extension/src/background/services/BackgroundScript.ts
@@ -658,7 +658,7 @@ export class BackgroundScript {
     }
 
     const receiver = stream.getPaymentPointer()
-    const {assetCode, assetScale, exchangeRate} = stream.getAssetDetails()
+    const { assetCode, assetScale, exchangeRate } = stream.getAssetDetails()
     const amount = Math.floor(1e9 * exchangeRate).toString() // 1 USD, assetScale = 9
 
     try {
--- a/packages/coil-extension/src/background/services/BackgroundScript.ts
+++ b/packages/coil-extension/src/background/services/BackgroundScript.ts
@@ -24,7 +24,7 @@ import {
   SetStreamControls,
   StartWebMonetization,
   ToBackgroundMessage,
-  Tip
+  TipSent
 } from '../../types/commands'
 import { LocalStorageProxy } from '../../types/storage'
 import { TabState } from '../../types/TabState'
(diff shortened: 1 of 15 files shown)


sublimator:
@sabinebertram 

If you run `yarn format` from the root of the repository (or from each individual package if that would be faster) and commit the changes the CI should pass

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -657,6 +658,8 @@ export class BackgroundScript {
     }
 
     const receiver = stream.getPaymentPointer()
+    const {assetCode, assetScale, exchangeRate} = stream.getAssetDetails()
+    const amount = Math.floor(1000000000 * exchangeRate).toString() // 1 USD, assetScale = 9
```
sublimator, line 662: Using e notation, e.g. `1e9`, is generally easier to read

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -674,6 +677,16 @@ export class BackgroundScript {
         }
       })
       this.log(`sendTip: sent tip to ${receiver}`, result)
+      const message: Tip = {
```
sublimator, line 680:
```
const message: SendTip = {
   command: 'sendTip'
```
trivial nit: I think SendTip is a bit more in alignment with the rest of the names used

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -657,6 +658,8 @@ export class BackgroundScript {
     }
 
     const receiver = stream.getPaymentPointer()
+    const {assetCode, assetScale, exchangeRate} = stream.getAssetDetails()
```
sublimator, line 661: Need to handle the rare case that assetCode/assetScale etc haven't been sent yet no?

packages/coil-extension/src/background/services/BackgroundScript.ts
sublimator: ah, already have that, so maybe `TipSent` ?

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -261,6 +269,9 @@ export class Stream extends EventEmitter {
       msSinceLastPacket: msSinceLastPacket,
       amount: data.amount.toString()
     })
+    this._assetCode = data.assetCode
```
sublimator, line 272: It might be better to use total sent and total delivered rather than an individual packet. I did note that there were some outlier packets where the exchangeRate can be considerably different than the average. 

I assume this may be later used for tipping UI showing destination currencies ? 

I see currently that tips are using a separate GraphQL mutation and are not really using the streaming method. 

Why not send the actual amounts sent from the tipping endpoint ? Is the endpoint async ?

packages/coil-extension/src/background/services/BackgroundScript.ts
sabinebertram: True, I have been wondering how. Maybe disable tipping if the stream connection cannot be established?

packages/coil-extension/src/background/services/Stream.ts
sabinebertram: @sharafian made me aware of the fact that the webmonetization standard defines the events with the destination currency and hence the tip event should also state values in the destination currency. We have been discussing whether we should change the standard now that STREAM receipts already contain the destination amounts. It may be desirable to show the user how much they sent instead of what the website received.

> Why not send the actual amounts sent from the tipping endpoint ? Is the endpoint async ?

I'm not sure I understand what you mean.

sublimator: > Why not send the actual amounts sent from the tipping endpoint ? Is the endpoint async ?
So it's a QraphQL mutation endpoint that does the heavy lifting of sending money. Does that endpoint return asynchronously or does it wait until the full send is complete. If the latter you should know exactly how much was sent (source) and received right ?

packages/coil-extension/src/background/services/BackgroundScript.ts
sublimator: That could work. You could do that on the popup side using existing state, hiding the tipping button if the balance is non-zero.

sublimator: s/if/unless

sublimator:
For the record, talked to @sharafian and he mentioned that at some point this event will contain some kind of STREAM receipt: 
https://github.com/interledger/rfcs/issues/568

Currently the event is primarily for client side, blind trust actions such as "Thanks for the tip!" popups.

PR #614, opened 2020-05-13
feat(coil-privacypass-sjcl): use node-x modules with browser shims

sublimator:


packages/coil-anonymous-tokens/src/portableFetch.ts
```diff
@@ -0,0 +1,18 @@
+// node-fetch will use window.fetch if in the browser
+// Avoids Failed to execute 'fetch' on 'Window': Illegal invocation TypeError:
+//   Failed to execute 'fetch' on 'Window': Illegal invocation
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
sublimator, line 4: Is this a TS 2.9 issue!?!?

sublimator: Doesn't seem to be ... odd

PR #673, opened 2020-05-27
refactor(web-monetization-demo): refactor and update README.md

sublimator:


sublimator:
Interesting! It's nice to have something self-contained (no redis etc).
Pondering removing or just linking to it in the readme ?

wilsonianb:
I was feeling compelled to remove the verifier here instead of having more or less the same implementation (to maintain) in two places.
(And the receipt-verifier shouldn't be moved to this monorepo, since can be used for ILP more generally than just web monetization.)

sublimator:
I think a toy example has some value (and I don't mind maintaining it), so I just added a link to the production verifier in the readme. Could actually do both, given the addition is just a dependency and altered script.

sublimator:
Ah, the balances and SPSP proxy apis are on different ports

PR #780, opened 2020-07-02
chore: rename main to master (#738)

sublimator:


sublimator:
Ugh, need to update my zsh functions and aliases

PR #781, opened 2020-07-02
feat(coil-extension): propagate logout to all contexts, incl private

sublimator:


sublimator:
@traviscrist @sharafian 

re: https://github.com/coilhq/coil/issues/3860

Chrome supports an extension instance per context, which when set makes everything work as expected with regards to authentication syncing and logging out. Unfortunately, this feature isn't implemented in FF.  

Should we disable the extension in private browsing contexts in FF or simply leave the behavior as is ?

https://developer.chrome.com/extensions/manifest/incognito#split
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/incognito

traviscrist:
> Should we disable the extension in private browsing contexts in FF or simply leave the behavior as is ?

I'd vote that we leave it as is, if we disable it then they can't use the extension for private browsing at all in FF. @sharafian thoughts?

sublimator:
@traviscrist 
That was the way I was leaning too

sharafian:
yeah I'd leave as-is for firefox. Will this mean that you need to sign in every time you open an incognito window in chrome, if you want to use WM?

sublimator:
> Will this mean that you need to sign in every time you open an incognito window in chrome
Yes, which could be annoying I guess.

sublimator:
@sharafian @traviscrist 
Well now that logout will carry across to all contexts maybe we should use that behavior for Chrome too. 
I don't really have any strong preference, given I rarely, if ever, use incognito/private sessions, but if "you need to sign in every time you open an incognito window in chrome" it could be quite annoying.

traviscrist:
I'm not sure if this is possible but I think the behavior should be:
1. If I am signed into Coil and I open a private window it keeps me signed in. (This seems useful and the best user experience)
1. If I log out of Coil in a private or not private window I stay logged out across both private and non private windows.

sublimator:
I think that’s how it’s implemented now if u want to try it

PR #930, opened 2020-08-24
feat(polyfill-utils): add more descriptive errors

sublimator:


packages/web-monetization-polyfill-utils/src/lib/MonetizationTagObserver.ts
```diff
@@ -3,7 +3,7 @@ import { v4 as uuid } from 'uuid'
 import { whenDocumentReady } from './whenDocumentReady'
 
 // eslint-disable-next-line @typescript-eslint/no-empty-function,@typescript-eslint/no-unused-vars
-const debug = (...args: unknown[]) => {}
+const debug = (..._: unknown[]) => {}
```
traviscrist, line 6: How does this work? Really just curious.. I don't think I've seen this pattern before.

sublimator: using `_` for unused variables is somewhat of a convention 
are you talking about the `...` syntax ? [rest params](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters)

essentially the `debug` function is a noop and you can change it to console.log while debugging
very "poor man's x" but it does the trick!

packages/web-monetization-polyfill-utils/src/lib/MonetizationTagObserver.ts
```diff
@@ -148,7 +148,12 @@ export class MonetizationTagObserver {
           this.onRemovedTag(tag)
         }
       } else {
-        throw new Error('<link> tag has been seen, ignoring <meta>')
+        throw new Error(
+          'Web-Monetization Error: ' +
+            'A <link rel="monetization"> tag has been seen, ' +
+            'so ignoring deprecated <meta name="monetization"> tag, ' +
```
traviscrist, line 154: Drop the `so` so its just `ignoring`

packages/web-monetization-polyfill-utils/src/lib/MonetizationTagObserver.ts
```diff
@@ -148,7 +148,12 @@ export class MonetizationTagObserver {
...
+            'use only monetization <link rel="monetization"> tags consistently'
```
traviscrist, line 155: `use` -> `using`

PR #1009, opened 2020-09-10
fix(coil-extension): tweak <link rel="monetization"> handling

sublimator:


packages/web-monetization-polyfill-utils/src/lib/resolvePaymentEndpoint.ts
```diff
@@ -1,9 +1,21 @@
...
+    : pointerOrUrl
 
   let url: URL
   try {
     url = new URL(httpUrl)
+    if (!url.protocol.match(/https?:/)) {
+      // noinspection ExceptionCaughtLocallyJS
+      throw new Error(
+        'SPSP endpoint must be specified as either a ' +
+          'payment pointer or fully resolved https:// url '
```
sublimator, line 16: http url, we allow non secure

PR #1019, opened 2020-09-11
feat(coil-extension): six debugging wip

sublimator:


packages/coil-extension/scripts/reinstall-six.sh
```diff
@@ -11,16 +11,16 @@ set -ex
...
 SIX_PKG=${SIX_PKG:-com.sec.android.app.sbrowser}
 
-# Uinstall old extension
-adb uninstall com.coil.android.six
-
-# Force stop Samsung Internet
-adb shell am force-stop $SIX_PKG
-# Clear Samsung Internet shared settings
-adb shell pm clear $SIX_PKG
+## Uinstall old extension
```
sublimator, line 14: Unsure if it's just the APK beta, but it seems no longer need to force stop the browser, clear the preferences and restart. Simply updating the extension APK seems to be picked up.

packages/coil-extension/src/background/services/utils.ts
```diff
@@ -61,3 +61,19 @@ export function createLogger(context: Context) {
     `color: white;`
   )
 }
+
+export function logLastError(from: string): () => void {
```
sublimator, line 65: This seems to log an error if the other side closes the channel immediately, which is quite often the case in unidirectional messaging.

PR #1040, opened 2020-09-15
feat(coil-extension): fast prepay tokens

sentientwaffle:
Instead of gradually paying out a token over 1 minute (via `AdaptiveBandwidth`), prepay entire tokens before each time period. If the user lingers on a page for some time, the tokens will be grouped together.

Description of the scheduling algorithm:

1. The user loads a monetized page.
2. A token is redeemed, and a stream is opened. The extension immediately pays 5 seconds worth of the token.
  - The immediate initial payment is important to ensure the page knows that the user agent is monetized.
  - Unlike subsequent payments, the first payment per page load is not paid ahead of time. This ensures that the user doesn't overpay when quickly navigating through many pages.
3. After 1 minute, the remainder of the initial token is spent. Immediately following, another full token is spent (for the following minute).
4. The extension continues to pay tokens:
  - Initially, it pays one token per minute.
  - After ~10 minutes (from page load), it switches to 2 tokens per 2 minutes.
  - After ~20 minutes (from page load), it switches to 3 tokens per 3 minutes.
  - And so on.

In other words, after the first minute the extension prepays _at most_ `1 minute + (10% of time on page).floor_to_nearest_minute` (the 10% is configurable -- I picked it pretty arbitrarily).

---

This PR depends on a couple of changes to coil's backend that have been merged, but not yet rolled out.

### DIFF
--- a/packages/coil-extension/src/background/services/BackgroundScript.ts
+++ b/packages/coil-extension/src/background/services/BackgroundScript.ts
@@ -396,11 +396,20 @@ export class BackgroundScript {
   ) {
     const tabState = this.tabStates.get(tabId)
 
+    const streamId = this.assoc.getStreamId({ tabId, frameId })
+    let isPaying = false
+    if (streamId) {
+      isPaying = this.streams.getStream(streamId).isPaying()
+    } else {
+      this.log('can not find top frame for tabId=%d', tabId)
+    }
+
     this.tabStates.setFrame(
       { tabId, frameId },
       {
         monetized: true,
-        total: total || (tabState.frameStates[frameId]?.total ?? 0)
+        total: total || (tabState.frameStates[frameId]?.total ?? 0),
+        isPaying
       }
     )
 
@@ -491,7 +500,12 @@ export class BackgroundScript {
 
     if (state) {
       const total = frameStates.reduce((acc, val) => acc + val.total, 0)
+      //const isPaying = frameStates.reduce((acc, val) => acc || val., false)
       this.storage.set('monetizedTotal', total)
+      this.storage.set(
+        'isPaying',
+        frameStates.reduce((acc, val) => acc || val.isPaying, false)
+      )
     }
     this.storage.set(
       'monetizedFavicon',
--- a/packages/coil-extension/src/background/services/BackgroundScript.ts
+++ b/packages/coil-extension/src/background/services/BackgroundScript.ts
@@ -500,12 +500,9 @@ export class BackgroundScript {
 
     if (state) {
       const total = frameStates.reduce((acc, val) => acc + val.total, 0)
-      //const isPaying = frameStates.reduce((acc, val) => acc || val., false)
+      const isPaying = frameStates.some(state => state.isPaying)
       this.storage.set('monetizedTotal', total)
-      this.storage.set(
-        'isPaying',
-        frameStates.reduce((acc, val) => acc || val.isPaying, false)
-      )
+      this.storage.set('isPaying', isPaying)
     }
     this.storage.set(
       'monetizedFavicon',
--- a/packages/coil-extension/src/background/services/PaymentScheduler.ts
+++ b/packages/coil-extension/src/background/services/PaymentScheduler.ts
@@ -18,8 +18,10 @@ const PREPAY = 0.1
 @injectable()
 export class PaymentScheduler {
   private sent = 0 // tokens sent
-  private startT = 0
-  private relT = 0 // time relative to first send
+  private startT = Date.now()
+  private stopT = 0
+  // time relative to start. Initially set in the future since the first (scheduled) token isn't until `startT+1m`.
+  private relT = TOKEN_DURATION
   private batchSize = 1 // tokens per batch
   private batchIndex = 0 // index within batch
   private timer?: NodeJS.Timer
@@ -28,11 +30,6 @@ export class PaymentScheduler {
   constructor(@logger('PaymentScheduler') private readonly debug: Logger) {}
 
   onSent(): void {
-    // The very first payment sent starts the scheduler's timer for the remaining
-    // points (just to simplify `timeOfSend`'s calculation).
-    if (this.sent === 0) {
-      this.startT = Date.now()
-    }
     this.sent++
     if (++this.batchIndex === this.batchSize) {
       this.batchIndex = 0
@@ -42,9 +39,16 @@ export class PaymentScheduler {
   }
 
   async wait(): Promise<void> {
+    // Don't pay when paused.
+    if (this.stopT) {
+      this.startT += Date.now() - this.stopT
+      this.stopT = 0
+    }
+
     const ms = this.waitTimeWithJitter()
     this.debug(
-      'waiting %dms before sending payment number %d',
+      'waiting until %s (%dms) before sending payment number %d',
+      new Date(Date.now() + ms),
       ms,
       this.sent + 1
     )
@@ -56,6 +60,7 @@ export class PaymentScheduler {
   }
 
   stop(): void {
+    this.stopT = Date.now()
     if (this.timer) clearTimeout(this.timer)
     if (this.onClose) this.onClose(new Error('closed'))
   }
--- a/packages/coil-extension/src/background/services/PaymentScheduler.ts
+++ b/packages/coil-extension/src/background/services/PaymentScheduler.ts
@@ -1,0 +1,78 @@
+import { injectable } from 'inversify'
+import { Logger, logger } from './utils'
+
+// The duration "paid" per token.
+const TOKEN_DURATION = 60_000 // milliseconds
+// For additional privacy, obfuscate token schedules by varying delays by ±jitter (uniform distribution).
+const JITTER = 5_000 // milliseconds
+// Except for the first token, full payment for interval [T₀,T₁] is sent at time T₀.
+// The interval is `tokensPerBatch * TOKEN_DURATION` milliseconds.
+// `tokensPerBatch` is the maximum of:
+// - `TOKEN_DURATION` (1 minute).
+// - `1+floor(sent*PREPAY)`
+//
+// For example, when `PREPAY==0.1`, after 10 minutes it will pay 2 tokens at a time.
+// After 20 minutes, it will pay 3 tokens at a time.
+const PREPAY = 0.1
+
+@injectable()
+export class PaymentScheduler {
+  private sent = 0 // tokens sent
+  private startT = 0
+  private relT = 0 // time relative to first send
+  private batchSize = 1 // tokens per batch
+  private batchIndex = 0 // index within batch
+  private timer?: NodeJS.Timer
+  private onClose?: (err: Error) => void
+
+  constructor(@logger('PaymentScheduler') private readonly debug: Logger) {}
+
+  onSent(): void {
+    // The very first payment sent starts the scheduler's timer for the remaining
+    // points (just to simplify `timeOfSend`'s calculation).
+    if (this.sent === 0) {
+      this.startT = Date.now()
+    }
+    this.sent++
+    if (++this.batchIndex === this.batchSize) {
+      this.batchIndex = 0
+      this.batchSize = 1 + Math.floor(this.sent * PREPAY)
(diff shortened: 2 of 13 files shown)


packages/coil-extension/src/popup/components/MonetizationAnimation.tsx
```diff
@@ -7,44 +7,16 @@ import { notNullOrUndef } from '../../util/nullables'
...
-  const [animated, setAnimated] = useState<boolean | null>(false)
-  const lastPacketRef = useRef(lastPacket)
-  lastPacketRef.current = lastPacket
+  const isPaying = localStorage.getItem('isPaying') === 'true'
+  const [animated, setAnimated] = useState<boolean>(isPaying)
 
   useEffect(() => {
-    let animateTimeout: number | null
+    const loopInterval = setInterval(() => {
+      setAnimated(localStorage.getItem('isPaying') === 'true')
```
sublimator, line 15: Can we access localStorage from the props?
We have a test page for the popup which shows it in various states(multiple instances of the popup component) and using a global wrecks that.

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -491,7 +500,12 @@ export class BackgroundScript {
 
     if (state) {
       const total = frameStates.reduce((acc, val) => acc + val.total, 0)
+      //const isPaying = frameStates.reduce((acc, val) => acc || val., false)
       this.storage.set('monetizedTotal', total)
+      this.storage.set(
+        'isPaying',
+        frameStates.reduce((acc, val) => acc || val.isPaying, false)
```
sublimator, line 507: inconsequential: can probably use frameStates.some(...)

packages/coil-extension/src/popup/services/PopupState.ts
```diff
@@ -25,6 +26,7 @@ export class PopupState implements PopupStateType {
   readonly monetized!: boolean
   readonly monetizedFavicon!: string
   readonly monetizedTotal!: number
+  readonly isPaying!: boolean
```
sublimator, line 29: aside: have been meaning to replace this with a Proxy

packages/coil-extension/src/background/services/PaymentScheduler.ts
```diff
@@ -0,0 +1,78 @@
...
+    const ms = this.waitTimeWithJitter()
+    this.debug(
+      'waiting %dms before sending payment number %d',
+      ms,
+      this.sent + 1
+    )
+    if (!ms) return
+    await new Promise((resolve, reject): void => {
+      this.timer = setTimeout(resolve, ms)
+      this.onClose = reject
```
sharafian, line 54: could there ever be more than one wait pending concurrently?

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -168,18 +166,29 @@ export class Stream extends EventEmitter {
       return
     }
 
-    // reset our timer when we start streaming.
-    const bandwidth = new AdaptiveBandwidth(
-      this._initiatingUrl,
-      this._tiers,
-      this._debug
-    )
+    const initialSendSeconds = 5
```
sharafian, line 169: nit, but this const could go at the top of the file

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -168,18 +166,29 @@ export class Stream extends EventEmitter {
...
+    const payScheduler = this.container.get(PaymentScheduler)
+    const bandwidth = new MaxBandwidth(0)
+    setTimeout(() => {
+      this._debug('finishing first payment')
+      bandwidth.sendMax = 2 ** 64
```
sharafian, line 174: nit; these consts also could be broken out

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -168,18 +166,29 @@ export class Stream extends EventEmitter {
...
 
     while (this._active) {
       let btpToken: string | undefined
       let plugin, attempt
+
       try {
-        btpToken = await this._anonTokens.getToken(this._authToken)
+        const redeemedToken = await this._anonTokens.getToken(this._authToken)
+        btpToken = redeemedToken.btpToken
+        // On page load, only send a small piece of the first token. After 1 minute, switch to full tokens.
```
sharafian, line 184: will this be called on page load, or every time the page becomes active (i.e. switching back into the tab?)

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -198,29 +207,40 @@ export class Stream extends EventEmitter {
...
           ilpReject.data.equals(await sha256(Buffer.from(btpToken)))
         ) {
-          this._debug('anonymous token exhausted; retrying, err=%s', e.message)
+          this._debug(
+            'anonymous token exhausted; retrying soon, err=%s',
+            e.message
+          )
           this._anonTokens.removeToken(btpToken)
+          payScheduler.onSent()
+          await payScheduler.wait()
```
sharafian, line 224: how come this wait only happens in the catch branch? is it because we run the try portion all the way until we exhaust the token?

sharafian (changes requested):
Overall looks good and I think I understand what's going on, but I've got a few questions just in case

sublimator:
Haven't really looked at this beyond a superficial skim, but a few questions:
1. Is this cutting down on radio "enough"?  The connections (WebSocket pings and whatever STREAM does) are still persistent right?
2. Is it possible to log 5 minutes of viewing over a certain amount of time between tab switches, but never actually pay out?

packages/coil-extension/src/background/services/Stream.ts
sentientwaffle: Yes. The `catch` block is always called -- either because of an actual error or (in the success case) because the token was exhausted.

packages/coil-extension/src/background/services/PaymentScheduler.ts
sentientwaffle: No, just one. It's called when a token hits `'exhausted capacity.'` (and tokens are sent sequentially, not in parallel).

packages/coil-extension/src/background/services/Stream.ts
sentientwaffle: While answering this question I realized that it works wrong right now, and found an unrelated bug. But how it should work (and will, once I fix it): switching back to a tab should start the schedule up right where it left off.

sentientwaffle:
1. After the first minute, tokens are paid in full. That means that the websocket/STREAM connections are reestablished for every token.
2. After the first minute, time is "prepaid" -- that is, the money is sent prior to the time period corresponding to the payment.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -121,9 +129,9 @@ export class Stream extends EventEmitter {
     this._requestId = requestId
     this._spspUrl = spspEndpoint
     this._authToken = token
-    this._tiers = container.get(BandwidthTiers)
     this._coilDomain = container.get(tokens.CoilDomain)
     this._anonTokens = container.get(AnonymousTokens)
+    this._schedule = container.get(PaymentScheduler)
```
sentientwaffle, line 134: `container.get(PaymentScheduler)` gets the same `PaymentScheduler` for every `Stream` instance instead of a different one every time. @sublimator do you know how to get `inversify` to make a new one?

sublimator: @sentientwaffle 
When configuring the container:
> container.bind(PaymentScheduler).toSelf().inTransientScope()

sublimator:
@sentientwaffle 
IIRC, when I was messing around with starting and stopping streams, Cinnamon would actually stop a video unless money had been recently streaming.

packages/coil-extension/src/background/services/Stream.ts
sentientwaffle: Thanks!

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -110,6 +108,8 @@ export class Stream extends EventEmitter {
     @logger('Stream')
     private readonly _debug: Logger,
     private container: Container,
+    @inject(PaymentScheduler)
```
sublimator, line 106: aside: You shouldn't need the inject() here afaik, but it doesn't really hurt

packages/coil-extension/src/background/services/PaymentScheduler.ts
sublimator: Related:
It's actually possible for multiple streams to be executing concurrently. The basic valid case is 3 Chrome windows open, all with tabs with a document.visibilityState of 'visible'

From memory this will use the same token for all 3 streams?

sublimator:
@sentientwaffle 
I built the extension to test it out

Load the event logger:
```
cd packages/coil-extension/test/fixtures
python -m http.server 4000
open http://localhost:4000/event-logger.html
```

When you switch away, it will fire a monetization stop event, then when you switch back it will emit a pending event, then the state will be "pending" for quite a long time until the next packet is sent.  

I guess this makes sense. 
I wonder if it might be pragmatic to add a started event immediately ?
I don't have a considered opinion yet myself, except to say the way it works now conforms with the existing model and is conceptually consistent.

I know for the Coil Blogs, we leaned upon a react [hasPaid](https://github.com/sharafian/react-web-monetization/blob/master/src/global.js#L41) hook to quickly show unlocked content before the state became "started".

sentientwaffle:
> When you switch away, it will fire a monetization stop event, then when you switch back it will emit a pending event, then the state will be "pending" for quite a long time until the next packet is sent.
> ...
> I wonder if it might be pragmatic to add a started event immediately ?

I think this is a good idea. Even though switching back (after switching away) doesn't immediately send money, it is _conceptually_ sending, since it has prepaid for the time.

sublimator:
> it is conceptually sending
Yeah, that's true

@sharafian 
What do you think? 

I'm not sure how these changes are going to effect everyone, but in principle, if we are going to break, it might be good to take the opportunity to make some wishlist changes you have up your sleeve ?

sharafian:
It definitely gets into the vaguer areas of the standard. Changing the schedule on which payout occurs shouldn't break the WM API, but it has the potential to break individual sites.

Some of that is our fault for using the monetization state, which isn't very flexible under models other than constant streaming. If we're talking about breaking changes I'd love to cut the monetization state model altogether but it would be a hard migration.

Here's a couple possiblitites, just brainstorming:

1. Make our changes adhere to the 'spirit' of the WM standard. They might technically be breaking if we're changing the semantics around monetization state, but the goal would be to minimize breakage among WM sites. Most sites have simple integrations and should require no upgrade. We could work with complicated sites like Cinnamon to make the necessary changes on their end if some breakage is unavoidable.

2. Go fully breaking (scrap the state model, force link tag, any other changes we want) and provide an upgrade path to sites (via a `version` attribute or something similar). We would keep the current functionality in the extension for sites that have not upgraded to the new standard. This would complicate our codebase somewhat but it would grant us the most flexibility to upgrade the WM standard in the future.

I don't love either option right now but there must be more ways that I haven't considered.

Option 1 would make WM harder to reason about because it has to 'lie' about what's happening behind the scenes in order to maintain compatibility. It could be a tough hole to dig ourselves out of in the future.

Option 2 would give us lots of flexibility but we'll have to keep supporting old WM versions for a long time unless we want to force sites to upgrade (very annoying dev experience). It also adds another dimension that would fragment docs and developer tools (if you're using multiple WM libraries on your site they'd need to be compatible versions)

What do you think @sentientwaffle @sublimator ? Maybe this conversation could be opened on the wicg/web-monetization issues to encourage wider discussion too.

sublimator:
I am not sure how/why "I" closed this. 
It was not intentional.

sublimator:
I'm not entirely, but for now I'd say we shouldn't rush this out.
Afaik, there isn't really any real-world pressing need for this PR? 

It might not be a bad idea to "allow" ourselves to think from scratch.
Think of an ideal, then a migration plan. 
If it doesn't seem feasible we could shoot it down.

PR #1216, opened 2020-10-23
feat(coil-extension): disabling by site/page/pp

sublimator:
## TODO
- [ ] Write a testing plan document
- [ ] See if "chrome.contextMenus" menus work with Samsung Internet, FF on Android 
- [x] Less "technical" sounding language ?
  - Domain -> Site, URL -> Page ?
  - [ ] Change "Payment Pointer" to recipient ?
    - image
- [ ] tool tips ?
- [ ] Persistence 
  - chrome.storage.sync 
     - [ ] Hash the entries ? 
     - [x] Need to set manifest.json key for a consistent local development extension id
        - the extension id is used for determining which remote StorageArea
        -  https://stackoverflow.com/a/46739698/11587288
     - works on FF and Chrome
       - Samsung? Safari ?
         - [x] Emailed Samsung @ sixsupport@samsung.com
            - "We are sorry that we don't support the API because of our internal policy. Please understand our policy." :(
         - [ ] Open Safari ticket?

## Testing

This section is first, simply to allow easy editing.

#### Test 2: Disable site, open site in new tab
Load a monetized page, with `localStorage.WM_DEBUG = true`
Use UI to disable site:
image

Open a new tab and reload the monetized site
Check the browser action icon shows disabled:
image

#### Test 1: Disable, refresh, enable 

Load a monetized site, with `localStorage.WM_DEBUG = true`
Use UI to disable site:
image
Check for a monetizationstop event logged:
image
Check the browser action icon shows disabled:
image

Refresh the page
Check that the stream does not start
Use UI to stop disabling the site
Check that the BrowserAction icon removes disabled badge: 
image
Check that the stream starts

## Disabling By Site/Page/Payment Pointer

The "This site is Web-Monetized" footer now shows disabled state, via toggling the $ to a prohibit icon.

(Note, these screenshots may be stale at this point, best to build and try it out)

Normal:
image

Cursor hovered over footer bar to preview click action (not shown: mouse cursor becomes a hand indicating a clickable):
image

Clicked with disabling menu showing:
image

It's a bit tricky with iframes, but the way it's currently implemented, if ANY of the disable toggles are checked (ORing), then all iframes in the tab will be disabled. The rules DO propagate to other tabs/frame but it's otherwise kept "simple".  For example if you block a.com by domain or url, then b.com (enabled) includes a.com as an iframe, then the a.com frame will be blocked. 

After disabling the current url, with the BrowserAction badge using a 🚫 emoji text with a dark background color:
image

Perhaps we could add a "one time exemption" button ? On a related note, with the current way the ORing of independent disabling controls works, there's no way to block a domain but enable for certain urls (exemptions) 

Other options:

We could use the contextMenus api:
image

#### Bug 1

CoilDev 0.0.51

For some reason the TabStates isn't book keeping isn't in sync with the BackgroundFramesService, such that there are FrameState's in existence in the former, but not the latter. The SetDisabling command calls `backgroundFramesService.getFrameNotNull(...)` which throws if the frame doesn't exist. 

image

This is hopefully fixed in CoilDev 0.0.51.1

#### Bug 2

There seems to be a regression in the youtube integration, with monetization not stopping when moving to an un-monetized video from a monetized one.


packages/coil-extension/src/background/services/BackgroundFramesService.ts
```diff
@@ -110,7 +111,7 @@ export class BackgroundFramesService extends EventEmitter {
     @logger('BackgroundFramesService')
     private log: Logger,
     @inject(tokens.WextApi)
-    private api: typeof window.chrome
+    private api = chrome
```
sublimator, line 114: hack to make intellij understand the types, silly regression it seems

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -215,45 +218,50 @@ export class BackgroundScript {
...
+      const { tabId, frameId } = event
+      const changedState = event.changed.state
       const isComplete = event.frame.state === 'complete'
-      const becameComplete = status === 'complete'
+      const becameComplete = changedState === 'complete'
+      const becameLoading = changedState === 'loading'
       const changedUrl = Boolean(event.changed.href)
 
-      // Always get the url from the tab
+      // Always get the current url
```
sublimator, line 233: change: always get the url from the full frame, not the bundle of "changed" properties, which may not include the url

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -215,45 +218,50 @@ export class BackgroundScript {
...
-            const ignored = this.api.runtime.lastError
-          }
-        )
+        this.checkAdaptedContent(from, tabId, frameId)
       }
     })
   }
 
+  private checkAdaptedContent(from: string, tabId: number, frameId: number) {
+    // Only extension apis can get callbacks for url change events
```
sublimator, line 252: aside: should filter this background script side to only check when the url is youtube/twitch

packages/coil-extension/src/background/services/DisablingService.ts
```diff
@@ -0,0 +1,212 @@
...
+    const urlDisabled = Boolean(this.disabled[href])
+    const domainDisabled = Boolean(this.disabled[origin])
+
+    if (isTopFrame) {
+      const disabling = this.tabStates.get(tab).disabling
+      disabling.disableDomain = domainDisabled
+      disabling.disableUrl = urlDisabled
+      this.tabStates.set(tab, { disabling })
+    }
+    // TODO: if disable rules apply, then apply enable or disable
```
sublimator, line 63: kill comment

packages/coil-extension/src/background/services/DisablingService.ts
```diff
@@ -0,0 +1,212 @@
...
+  /**
+   * Upon changes, must check all open tabs/frames to see if they apply and
+   * update the tab state accordingly.
+   *
+   * If ANY of the controls are set to disabled, disable ALL streams for the
+   * active tab, then apply the rules to other tabs (which may be backgrounded).
+   *
+   */
+  setDisabled(activeTab: number, controls: DisablingControls) {
+    console.log(
```
sublimator, line 92: add a @logger()

packages/coil-extension/src/background/services/DisablingService.ts
```diff
@@ -0,0 +1,212 @@
...
+      href: string
+      paymentPointer: string | null
+    }> = []
+    for (const tabId of this.tabStates.tabKeys()) {
+      const tabState = this.tabStates.get(tabId)
+      for (const frameId of Object.keys(tabState.frameStates).map(Number)) {
+        const frame: FrameSpec = { frameId, tabId }
+        const state = tabState.frameStates[frameId]
+        if (!this.framesService.getFrame(frame)) {
+          console.log('getFrameStates missing frame', JSON.stringify(frame))
```
sublimator, line 160: TODO: not sure why we are getting this 
Potentially the tabStates frame bookkeeping isn't 100%
No ... 
It was due to stopWebMonetization deleting the frameState

packages/coil-extension/src/background/services/DisablingService.ts
```diff
@@ -0,0 +1,212 @@
...
+      return true
+    }
+    const paymentPointer = this.tabStates.getFrameOrDefault(frame)
+      .monetizationDetails?.paymentPointer
+
+    const urlObject = new URL(
+      notNullOrUndef(this.framesService.getFrame(frame)?.href)
+    )
+    const { href, origin } = urlObject
+    // TODO: normalize
```
sublimator, line 201: Yes, TODO, normalize payment pointers to urls

packages/coil-extension/src/background/services/Streams.ts
```diff
@@ -68,4 +68,8 @@ export class Streams extends EventEmitter {
   getStream(id: string) {
     return this._streams[id]
   }
+
+  getStreams() {
```
sublimator, line 72: Remove: unused

packages/coil-extension/src/background/services/TabStates.ts
```diff
@@ -58,16 +58,24 @@ export class TabStates {
...
     return state
   }
 
-  private makeFrameStateDefault() {
+  private makeFrameStateDefault(): FrameState {
     return {
       monetized: false,
+      monetizationDetails: null,
+      disabled: false,
+      interactive: false,
```
sublimator, line 78: the pauseWebMonetization script from the background on idle, on visibilityState === 'hidden', would set this to false

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -321,18 +329,29 @@ export class BackgroundScript {
...
+          // this.tabStates.setIcon(tabId, 'inactive')
           this.tabStates.setIcon(tabId, 'monetized')
+
           if (hasBeenPaid) {
-            const state =
-              tabState.playState === 'playing'
-                ? 'streaming'
-                : 'streaming-paused'
-            this.tabStates.setIcon(tabId, state)
+            this.tabStates.setIcon(tabId, 'streaming')
```
sublimator, line 340: Note that the state of the icon is actually a combination of these setIcon calls. It's probably not very obvious to those uninitiated

sublimator:
Discussed with @fruehle 

> Maybe we can replace the “This site is Web-Monetized” with”“Web Monetization disabled”
> Can use `–` instead of banned emoji:
> The en dash () is slightly wider than the hyphen (-) but narrower than the em dash (—).

fruehle:
Some comments:
- Could we change the color of the active toggle to red?
- Change the copy next to the 🚫 emoji from "This site is Web-Monetized" to "Web Monetization disabled"

packages/coil-extension/src/background/services/BackgroundScript.ts
ChRriiZz:
````

Issue #1248, opened 2020-10-31
Coil extension Content Security Policy violation prevents loading


wmurphyrd:
Hi It's Will from Immers Space, one of the Grant for the Web flagships. 

When I try to use Firefox Coil add-on with a production site using Content Security Policy, the `document.monetization` polyfill fails to load and the Coil add-on generates a CSP console error:

```
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”). content.js:21:461461
```

This error can be observed here: https://hub.medieval-castle.club/

Here's some [documentation from Mozilla on the issue wrt addons](https://blog.mozilla.org/addons/2019/12/12/test-the-new-csp-for-content-scripts/). I can open up CSP to inline sources to get it working now, but I'd rather not leave that vulnerability open long-term.

sublimator:
Thanks for the report, will look into it in the coming days

wmurphyrd:
I was able to get it working at the above url by adding the following 4 script hashes to my `script-src` directive:

```
'sha256-W5yaJ6UM3/kOJa12aRVSLOEOKdAUYAWZPM1bUuaTJYQ=' 
'sha256-XpyxuqRQmj1o8ovYZlIA71UXSYTvYdV8kOb55p+lrNo='
'sha256-tie542PGbiDGOm9MefVIzDBZf4Nt5wTagAHT/BKEB94='
'sha256-9QLzkf1LE5s0CtnpqUvwkWr7DV4GRfQLGt/tFNT19h0='
```

This covers the initial polyfill script, and the 3 versions of the `monetization.state` update script. I think adding these to the add-on manifest under content_security_policy.content_scripts may solve the issue everywhere

Since the initial link is now fixed, here's a new one to obvserve the error https://nebulous-confused-twig.glitch.me/ ([source](https://glitch.com/edit/#!/nebulous-confused-twig?path=server.js%3A41%3A33))

sublimator:
Thanks for the update

sublimator:
I've had a quick look at this link and it seems that this is referring to manifest v3, rather than v2?  
https://blog.mozilla.org/addons/2019/12/12/test-the-new-csp-for-content-scripts/

I experimented with the new "content_security_policy" as an object, didn't really see any effect, then tried converting to manifest version 3, but was unable to load the add-on:
![image](https://user-images.githubusercontent.com/525211/97829333-5480de00-1cfc-11eb-9db4-9d8f7ba7e54a.png)

[...]

PR #1254, opened 2020-11-03
fix(coil-extension): reduce code injections to 1

sublimator:
Related to #1248 

We may end up moving to emitting events on window directly, but even in that case, we'll need to provide support for old sites, along with deprecation warnings, so there is some value in reducing the amount of injections.


wmurphyrd:
If you were to wrap that script injection in a try/catch, I think it would it be now possible to have full functioning under CSP without the script-src directives by simply running that short  `document.monetization` polyfill script from the main paige

sublimator:
@wmurphyrd  Thanks for your input!

Yeah, I was pondering the same thing actually. 
It's potentially more maintainable that way too.

Issue #1268, opened 2020-11-06
Update compilerOptions.target to 2020


sublimator:
> https://node.green/
Seems es2020 is supported for node 14

What about Chrome/FF ? 
Assumedly as well

PR #1354, opened 2020-12-02
feat(coil-extension): replace ?? with ||

winstonchen:
error caused by ?? - || can be used instead

error when building crx with build.sh/package.sh. Not completely familiar with typescript.

### DIFF
--- a/packages/coil-webpack-utils/src/index.ts
+++ b/packages/coil-webpack-utils/src/index.ts
@@ -17,7 +17,7 @@ export const configureNodePolyfills = (wpConf: webpack.Configuration) => {
     stream: require.resolve('stream-browserify/')
   }
   wpConf.plugins = [
-    ...(wpConf.plugins ?? []),
+    ...(wpConf.plugins || []),
     new webpack.ProvidePlugin({
       process: ['process']
     }),


sublimator:
@winstonchen 

I am not sure what's going on here, but we have the CI set up compiling all the TypeScript every PR and merge to main. 
Therefore, in general, this shouldn't be a problem. Perhaps it's specific to your local setup or the `nd-six-support-wip-2020-09-11` branch ? I will check the state of the latter.

Can you show me the actual error messages you saw ?

sublimator:
@winstonchen 

I just recently updated the TypeScript compile target (the JavaScript that emitted) to es2020 in #1268 

I just confirmed that `??` will be emitted as is: 

The typescript here:
```typescript
export function testTheory(a: unknown) {
  return a ?? '<<<-- Do we keep the `??` in es2020 target ?'
}
```

Will be compiled as the following JavaScript:
```javascript
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.testTheory = void 0;
function testTheory(a) {
    return a ?? '<<<-- Do we keep the `??` in es2020 target ?';
}
exports.testTheory = testTheory;
//# sourceMappingURL=checker.js.map%           
```

Did the error occur at runtime ?

sublimator:
```
does anyone know how to get the staging extension to work? i keep getting the following webpack error
[webpack-cli] /Users/alan/projects/web-monetization-projects/packages/coil-webpack-utils/src/index.ts:41
        ...(wpConf.plugins ?? []),
                            ^
SyntaxError: Unexpected token ?
```

From our dev channel recently. 

@winstonchen 

What version of node are you running ? 
The CI is setup to use node v14 and that's what I use locally

winstonchen:
Ahh,  I was using an older node version (v11). Updating to v14 resolved this issue.

sublimator:
> Updating to v14 resolved this issue.

Good to hear!

I had configured "engines", at least in the root package.json:
https://github.com/coilhq/web-monetization-projects/blob/main/package.json#L100-L102

So I am not sure why yarn wasn't complaining ??

winstonchen:
I also tried building on a clean Ubuntu vm install, same issue. I guess the default node install is not latest even after update.

sublimator:
@winstonchen 

BTW I found you could use chrome to sign a crx, then jarsigner to rebuild the apk, followed by a call to adb to reinstall.
There are some scripts in the repo to support that. 

See:
https://github.com/coilhq/web-monetization-projects/blob/nd-six-support-wip-2020-09-11/packages/coil-extension/scripts/reinstall-six.sh
https://github.com/coilhq/web-monetization-projects/blob/nd-six-support-wip-2020-09-11/packages/coil-extension/scripts/sixpacker.sh

The following will rebuild the APK
```
export AFTER_EMIT_SHELL_CMD="./scripts/reinstall-six.sh"      
yarn dev-chrome-prod
```

I'm not sure if you can just create a new set of keys and use them. It may only be required that the keys are the same upon reinstalls/upgrades ?

Just open issues if you have any questions

Thanks!

winstonchen:
You can use different keys to sign the APK, however reinstalling would require same key signing. Reinstall checks if signatures matches.

I will take a look at these scripts.

PR #1385, opened 2020-12-08
feat(coil-extension): add GPCService sketch

sublimator:
I had a look at the list of extensions at https://globalprivacycontrol.org/.

Two things struck me:
* Most of them do extracurricular activities too, blocking ads/"trackers" etc
* I looked at a few in detail and they request a whole raft of permissions
  * No qualms using the blocking apis - performance can't be that bad

It seems like the sort of thing that the average Coil user would most likely be happy to opt-out of, rather than opt-in, so:
* just send the header by default
  * opt-out setting in the extension
  * add to coil.com unified settings page as time permits

- [x] Try [declarativeNetRequest](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/):
  - https://github.com/coilhq/web-monetization-projects/pull/1659

### DIFF
--- a/packages/coil-extension/manifest.json
+++ b/packages/coil-extension/manifest.json
@@ -32,5 +32,5 @@
       "run_at": "document_start"
     }
   ],
-  "permissions": [ "webNavigation", "<all_urls>" ]
+  "permissions": [ "webNavigation", "<all_urls>", "webRequest", "webRequestBlocking" ]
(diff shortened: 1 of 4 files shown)


sublimator:
Try this:
https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/

PR #1390, opened 2020-12-09
extension: batch post-payment of tokens

sentientwaffle:
Supersedes https://github.com/coilhq/web-monetization-projects/pull/1040, extension side of https://github.com/coilhq/coil/pull/4712 (older reverted PR: https://github.com/coilhq/coil/pull/4582)

#### Batch info
- The very first payment (in both the extension and oauth-scripts) is paid gradually -- 5s, 5s, 20s, 30s (see: `firstMinuteBandwidth`).
  - This ensures that if the user is on a page for only a short time (e.g. before navigating away or closing a tab) a correspondingly small amount is paid.
- The batch/timer logic is in `PaymentScheduler.ts`.
- Time while the tab is activated is tracked. Initially `PaymentScheduler` will pay one token per minute, when enough time on the page has accumulated. After 10 minutes, it pays 2 tokens every 2 minutes (and so on, to a maximum of 5tokens/5mins).
- Extension pays at the end of an interval, oauth-scripts pays for a period of time beforehand.

#### Extension manual tests
- Closing a tab: When a tab is closed, the extension should send a "catch-up" payment of one or more tokens (the last one potentially being partial).
- Deactivating a tab: When a tab is deactivated, the extension doesn't send a catch-up payment. If after being deactivated, the tab is:
  - closed: A catch-up payment is sent for any unpaid time.
  - activated: The tab will resume its timer, and send payment when it has accumulated enough active time to merit a batch payment.
- Quickly switching through different tabs: afaik this is just a special case of "Deactivating a tab". Payments are only sent when batches are ready.
- Flicking back and forth between tabs: (see above)

#### Oauth-scripts manual tests:
- Closing a tab: n/a; the oauth-script can't run in a closed tab. Since oauth-scripts prepays, this is not an issue.
- Deactivating a tab: When a tab is deactivated, the extension doesn't send a catch-up payment. If after being deactivated, the tab is:
  - closed: (see above)
  - activated: The tab will start consuming its "prepaid time", and eventually pay another batch when its current prepaid time has run out. The "prepaid time" is not consumed while the tab is deactivated.

## State coherence violations

This will emit progress events in background tabs while they are in the "stopped" state. 
It's very likely not a big deal and we are moving towards deprecating "state" and non progress events. 

#### Solutions

- For the situation of a tab in the stopped state could we do our hack of changing state to 'started', emitting the start event, emitting the progress event, and then going back to 'stopped'? (Ben)
- Queue the event until tab reactivation ?
- None - just live with it

## Packet "loss" Issues (circa receipt submission)

It's always been the case that we just discard packets in flight and don't report them to a deactivated tab.

The amounts are so small that it hasn't really mattered (has been my reading anyway)
But in a post pay world where you're potentially only paying once every 5 minutes (illustrative) ?
The amounts would be much larger.

Do you just report a `monetizationprogress` when the tab is deactivated? even though streaming is "stopped" (noting that document.monetization.state will be deprecated) and the tab isn't active (a more durable strangeness)
Queue the event until tab reactivation?

Another instance of packet loss is when closing a tab. Because it's "post" paid, it is necessary to do catch-up payments, but of course you have no tab to report the payments to. This has implications for client side submission of receipts for keeping a balance.

#### Solutions

- If the payment pointer on the page were generated per user (or included session info), then payment sent even after the page is closed could be counted towards future sessions (Ben) 
- Could revive the "deprecated" Web-Monetization-ID which is sent in the SPSP requests
- Standardize receipt submission as part of the WM standard (not appealing at present, but listed for documentation sake)

TODO before merging:
- [x] More testing
   - [ ] (dj) write manual testing steps to release-checklist.md
- [x] ~Add batched pre-pay to the oauth script.~
- [x] Figure out the missing `outgoing_money` event bug.
      - https://github.com/coilhq/web-monetization-projects/pull/1390#issuecomment-748134302
- [x] (dj) Add oauth fast-payment tokens to the server side.
- [x] (dj) Add oauth fast-payment tokens to the oauth-scripts.
- [x] (dj) Test start event fires properly in oauth-scripts
   - Run `yarn serve-it ../coil-extension/test/fixtures/event-logger.html` from `packages/coil-oauth-scripts`
   - Run `OUTPUT_TO=$PWD/../coil-extension/test/fixtures yarn build --watch` from `packages/coil-oauth-scripts`
   - open http://localhost:4000 in browser and set localStorage.WM_DEBUG = true in devtools
- [x] (ext) fires start event even when page is backgrounded
   -  image 
   - Check "using extension" (chrome://extensions to enable/disable) 
   - image
   - load event-logger.html via `yarn serve-it` and enabled logging as above
   - deactivate tab quickly after refreshing
- [x] (ext) Start event not firing:
   - https://github.com/coilhq/web-monetization-projects/pull/1390#issuecomment-764369419
   - reported by @sublimator, reprod by @wilsonianb (ubuntu 18.04, chrome 88.0.4324.96 (Official Build) (64-bit))
   - DocumentMonetization#setState currently only fires monetizationstop/monetizationpending events: https://github.com/coilhq/web-monetization-projects/blob/main/packages/web-monetization-wext/src/content/DocumentMonetization.ts#L87-L96
 - [x] (ext) Unexpected state bug
    ``` 
    Error in event handler: Error: expecting state transition
    at DocumentMonetization.postMonetizationStartWindowMessageAndSetMonetizationState (chrome-extension://hcohoecolgmlofifjaobjhidpoaciknp/content.js:41031:19)
    at chrome-extension://hcohoecolgmlofifjaobjhidpoaciknp/content.js:1071:35
    ```
    - https://github.com/coilhq/web-monetization-projects/blob/1c9696b7dc5e2520c0f5f7075c947655e4b84bed/packages/coil-extension/src/background/services/BackgroundScript.ts#L262-L274
    - It's possible that the code above could run AFTER a resume has sent `sendSetMonetizationStateMessage`
      - due to packetNumber not being the packet number of the Stream, but instead the StreamAttempt (or what that was refactored to)


packages/coil-extension/src/util/debounce.ts
```diff
@@ -0,0 +1,23 @@
...
+export function debounce(fn: F, wait: number): F {
+  let args: any[] | null = null
+  let timer: NodeJS.Timer | null = null
+
+  function flush(): void {
+    if (args !== null) fn(...args)
+    args = timer = null
+  }
+
+  return function (...rest): void {
```
sublimator, line 14: How is (...rest) getting past `"noImplicitAny": true` in the tsconfig ??

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -712,7 +716,10 @@ export class BackgroundScript {
     const id = this.assoc.getStreamId(frame)
     if (id) {
       this.log('resuming stream', id)
-      this.sendSetMonetizationStateMessage(frame, 'pending')
+      this.sendSetMonetizationStateMessage(
+        frame,
+        this.hasStreamPaid(frame) ? 'started' : 'pending'
```
sublimator, line 721: So we decided with sending started immediately in this case

packages/coil-extension/src/util/debounce.ts
sentientwaffle: I don't know. Maybe since it is `any[]`, not `any`?

packages/coil-extension/src/background/services/BackgroundScript.ts
sentientwaffle: Yes, I think that's best. It is resuming payment, and has successfully paid already.

sublimator:
The puppeteer->playwright tests have failed a few times 
Will need to look into it

sublimator:
> Figure out the missing outgoing_money event bug.

What's that ?

sentientwaffle:
Yeah, I'm not sure what's going on with the puppeteer/playwright test.

---

> Figure out the missing outgoing_money event bug.

This is a bug (probably in `ilp-protocol-stream`, though that is just a theory atm) I've noticed while testing this PR (the extension part) -- though I think it predates this branch.

Steps:

1. The extension has a full token it wants to pay.
2. Open a stream connection, and set the `sendMax` to `2**64`.
3. The stream pays, then closes when it receives the `exhausted capacity` error from flat-stacks.
4. On the extension side, `stream.totalSent` and `connection.totalSent` occasionally report less than the full token value.
    - The timing of these is such that it isn't just partial tokens being reused from the cache.
    - It tends to be only a little less (I estimate one `StreamMoney` frame short). But sometimes it is much less.
    - The `totalSent` stays the same even after pausing before checking it.
    - My guess is that there is some race condition, and the connection is sometimes closed before the `StreamMoney` frame is either sent by flat-stacks' Stream or processed by the extension's Stream.

sublimator:
I'm extending holidays after a health scare. 
Let me know if you would like to publish the extension to [CoilDev](https://chrome.google.com/webstore/detail/coildev/hcohoecolgmlofifjaobjhidpoaciknp) for people (Coilers/Cinnamon/etc) to tire kick.

wilsonianb:
I tested the CoilDev :point_up_2: extension with a [modified web-monetization-demo](https://github.com/wilsonianb/web-monetization-projects/commit/fbb96fa26290fe6974d1bdb5906d6777967f7be5) with the following results:
```
+0s 1679609
+1m10s 20175534
+2m6s 20182842
+3m8s 20183586
+4m7s 20056044
+5m8s 20041042
+6m12s 20057412
+7m7s 20023044
+8m11s 20008140
+9m8s 19921602
+10m12s 19852098
+12m7s 39643950
+14m6s 39715236
+16m9s 39743208
+18m9s 39744940
+20m8s 39787092
+23m8s 59740452
+26m5s 59830480
+29m7s 60028920
```

packages/coil-oauth-scripts/src/Stream.ts
```diff
@@ -319,4 +332,30 @@ export class Stream {
   pause() {
     return this.stop(false)
   }
+
+  // Returns token fraction.
+  private currentStreamSendMax(): number {
+    const elapsed = this.schedule['watch'].time()
```
sublimator, line 338: Maybe can expose something public on schedule to get this data with private['investigations'] ?

packages/coil-oauth-scripts/src/Stream.ts
```diff
@@ -226,22 +241,19 @@ export class Stream {
...
         const onUpdateAmountInterval = async () => {
-          const sendMaxAmount = await adaptiveBandwidth.getStreamSendMax()
-          if (stream.isOpen()) {
-            stream.setSendMax(sendMaxAmount)
-          }
+          // Round to ignore floating point errors.
+          const newSendMax = Math.round(
+            throughput * 60 * this.currentStreamSendMax()
+          )
+          if (+stream.sendMax < newSendMax) stream.setSendMax(newSendMax)
```
sublimator, line 248: We generally prefer Number(string) in the rest of the codebase, but it's not a biggy.

sublimator:
If you go to a web-monetized page, then set localStorage.WM_DEBUG = true, the content script will log WM events:
image

I was testing loading a page, switching the active tab to see a stop event, then switching back expecting to see a start event. Unfortunately no start event seen, and I'm actually seeing an error:
image

sentientwaffle:
Can you elaborate on that issue? I just tried what you described locally and it worked w/o issue.

A couple questions:

- When you switched the active tab `A -> B -> A`, is tab `B` monetized?
- Does the error happen consistently or intermittently?

packages/coil-oauth-scripts/src/Stream.ts
sentientwaffle: Yeah, I prefer `+string` since its more succinct.

sublimator:
@sentientwaffle 

I didn't really look much further upon seeing the bug, given the PR is still in draft mode, the "more testing" checklist item is unchecked and I had some other work to attend to. I got the impression it wasn't really finished. 

> I just tried what you described locally 

What do you mean by "locally" ? Connecting to a local backend ?

sublimator:
> Does the error happen consistently or intermittently?

Well, I'm assuming intermittently if it works for you, but as fate would have it, it popped up first time I thought of testing that case. So I guess it's probably lurking there somewhere even if for you it "works for me" ?

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -314,212 +319,138 @@ export class Stream extends EventEmitter {
...
+    this.state = StreamLoopState.Loop
+    if (this.running) return // another 'run' loop is already active
+    this.running = true
+    this.schedule.start()
+    let attempts = 0 // count of retry attempts when state=Ending
+    while (this.state !== StreamLoopState.Done) {
+      let tokenPiece = 1.0
+      switch (+this.state) {
+        case StreamLoopState.Done:
+          continue // stop looping (unreachable)
```
sublimator, line 352: unreachable due to while($condition)

packages/coil-anonymous-tokens/src/index.ts
```diff
@@ -124,22 +129,23 @@ export class AnonymousTokens {
...
+  ): Promise<RedeemedToken | undefined> {
     const usableToken = deserializeToken(token)
     const redeemRequest = BuildRedeemHeader(usableToken, '', '')
     const response = await portableFetch(this.redeemerUrl + '/redeem', {
       method: 'POST',
       headers: { 'content-type': 'application/json' },
       body: JSON.stringify({
-        bl_sig_req: redeemRequest
+        bl_sig_req: redeemRequest,
+        mode: 'fast'
```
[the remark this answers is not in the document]
sentientwaffle: `mode:"fast"` is a new option in the redeemer service -- when `fast` is specified the new BTP token will have refill=0 and capacity=aggregate. This ensures that the token can be spent rapidly within just a few packets, rather than gradually over a full minute.

packages/coil-anonymous-tokens/test/jest/redeemer.test.ts
```diff
@@ -40,7 +41,7 @@ const TOKENS_SERVICE_TEST = Boolean(process.env.TOKENS_SERVICE_TEST)
       await tokens['populateTokens'](coilAuthToken)
 
       const btpToken = await tokens.getToken(coilAuthToken)
-      const decoded = jwt.verify(btpToken, BTP_SECRET) as any
+      const decoded = jwt.verify(btpToken.btpToken, BTP_SECRET) as any
```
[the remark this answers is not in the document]
sentientwaffle: I don't see a variable called `redeemed` here.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -314,212 +319,138 @@ export class Stream extends EventEmitter {
...
-      const onMoney = (sentAmount: string) => {
-        // Wait until `setImmediate` so that `connection.totalDelivered` has been updated.
-        const receipt = this._ilpStream.receipt
-          ? this._ilpStream.receipt.toString('base64')
-          : undefined
-        setImmediate(this.onMoney.bind(this), sentAmount, receipt)
+  async run(
+    attempt: (tokenFraction: number) => Promise<Connection>
+  ): Promise<void> {
+    this.state = StreamLoopState.Loop
```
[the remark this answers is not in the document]
sentientwaffle: Even with `as StreamLoopState` on all three of those sets, compiling fails with:

```
src/background/services/Stream.ts:381:14 - error TS2678: Type 'StreamLoopState.Done' is not comparable to type 'StreamLoopState.Loop | StreamLoopState.Ending'.
381         case StreamLoopState.Done:
                 ~~~~~~~~~~~~~~~~~~~~
src/background/services/Stream.ts:411:16 - error TS2678: Type 'StreamLoopState.Done' is not comparable to type 'StreamLoopState.Loop | StreamLoopState.Ending'.
411           case StreamLoopState.Done:
                   ~~~~~~~~~~~~~~~~~~~~
```

I think the issue is the narrowing in the `while (this.state !== StreamLoopState.Done) {` comparison. Admittedly, the comparison on 381 is unnecessary, but the one on 411 is important.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -314,212 +319,138 @@ export class Stream extends EventEmitter {
...
+          continue // stop looping (unreachable)
+        case StreamLoopState.Loop: // keep going; only pay full tokens
+          // After the very first token payment, switch to post-payment.
+          if (0 < this.schedule.totalSent()) {
+            const hasToken = await this.schedule.awaitFullToken()
+            if (!hasToken) continue // the wait was interrupted, likely because the stream was paused/stopped
+          }
+          break
+        case StreamLoopState.Ending:
+          tokenPiece = Math.min(this.schedule.unpaidTokens(), 1.0)
```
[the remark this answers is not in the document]
sentientwaffle: > after having prepaid a minute

Do you mean having paid/postpaid a minute? The extension only **pre**pays `INITIAL_SEND_SECONDS` (5 seconds), everything after that is post-payment.

When a stream is closed at 1.5 minutes, and 1 minute has already been paid, there is 30 seconds left to pay. So `this.schedule.unpaidTokens()` is 0.5, and it will pay half a token before ending.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -314,212 +319,138 @@ export class Stream extends EventEmitter {
...
-
-    return new Promise((resolve, reject) => {
-      const onMoney = (sentAmount: string) => {
-        // Wait until `setImmediate` so that `connection.totalDelivered` has been updated.
-        const receipt = this._ilpStream.receipt
-          ? this._ilpStream.receipt.toString('base64')
-          : undefined
-        setImmediate(this.onMoney.bind(this), sentAmount, receipt)
+  async run(
+    attempt: (tokenFraction: number) => Promise<Connection>
```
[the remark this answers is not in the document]
sentientwaffle: The extension pays in terms of tokens, regardless of the tokens' underlying throughput. As far as the payment scheduler is concerned, a token issue at <$4.50 is identical to one issued at >$4.50.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -314,212 +319,138 @@ export class Stream extends EventEmitter {
...
-      sourceAmount: sentAmount,
-      sourceAssetCode: this._connection.sourceAssetCode,
-      sourceAssetScale: this._connection.sourceAssetScale
-    })
+async function firstMinuteBandwidth(
+  stream: DataAndMoneyStream,
+  throughput: number
+): Promise<void> {
+  // Start paying immediately so the page knows that the user agent is monetized.
+  // But the amount it isn't too high so that the user doesn't deplete their
```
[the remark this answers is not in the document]
sentientwaffle: I'll clarify the comment.

packages/coil-oauth-scripts/src/Stream.ts
```diff
@@ -7,23 +7,25 @@ import * as IlpStream from 'ilp-protocol-stream'
...
   MonetizationProgressEvent,
   MonetizationStartEvent
 } from '@web-monetization/types'
-import { BandwidthTiers } from '@coil/polyfill-utils'
 
 import { DocumentMonetization } from './DocumentMonetization'
 import { debug } from './logging'
 
 const UPDATE_AMOUNT_INTERVAL = 2000
+// Use a high enough timeout to allow for large intervals between batches of payments.
```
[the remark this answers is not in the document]
sentientwaffle: No, the extension opens a new stream connection for every token sent, so they are never open for long.
In contrast, oauth-scripts keeps one connection open, and increases the `sendMax` for every token sent.

packages/coil-extension/src/background/services/Stream.ts
sentientwaffle: Yeah, its unreachable. I included the case anyway for clarity.

packages/coil-extension/src/background/services/Stream.ts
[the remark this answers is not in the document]
sublimator: Right, yah. So this will only ever pay at most one token' worth when winding down? Would there ever be situations where the site is "owed" more than that?

packages/coil-extension/src/background/services/Stream.ts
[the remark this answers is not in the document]
sublimator:
```
diff --git a/packages/coil-extension/src/background/services/StreamLoop.ts b/packages/coil-extension/src/background/services/StreamLoop.ts
index 6d73f903..d0e3fd5f 100644
--- a/packages/coil-extension/src/background/services/StreamLoop.ts
+++ b/packages/coil-extension/src/background/services/StreamLoop.ts
@@ -30,16 +30,16 @@ export class StreamLoop extends EventEmitter {
   async run(
     attempt: (tokenFraction: number) => Promise<Connection>
   ): Promise<void> {
-    this.state = StreamLoopState.Loop
+    this.state = StreamLoopState.Loop as StreamLoopState
     if (this.running) return // another 'run' loop is already active
     this.running = true
     this.schedule.start()
     let attempts = 0 // count of retry attempts when state=Ending
     while (this.state !== StreamLoopState.Done) {
       let tokenPiece = 1.0
-      switch (+this.state) {
-        case StreamLoopState.Done:
-          continue // stop looping (unreachable)
+      switch (this.state) {
+        // case StreamLoopState.Done:
+        //   continue // stop looping (unreachable)
         case StreamLoopState.Loop: // keep going; only pay full tokens
           // After the very first token payment, switch to post-payment.
           if (0 < this.schedule.totalSent()) {
@@ -67,9 +67,9 @@ export class StreamLoop extends EventEmitter {
         attempts = 0
       } catch (err) {
         this.emit('run:error', err)
-        switch (+this.state) {
-          case StreamLoopState.Done:
-            break
+        switch (this.state) {
+          // case StreamLoopState.Done:
+          //   break
           case StreamLoopState.Ending:
             if (++attempts === STOP_RETRIES) {
               this.debug('too many errors; giving up')
```

What's wrong with that?
You can always leave the commented out cases there for "clarification" 
What's happening on the previous "line 411" ?

packages/coil-anonymous-tokens/test/jest/redeemer.test.ts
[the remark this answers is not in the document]
sublimator: exactly ? 

`btpToken.btpToken`

packages/coil-extension/src/background/services/Stream.ts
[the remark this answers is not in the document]
sentientwaffle: > So this will only ever pay at most one token' worth when winding down?

No. The `StreamLoop` will keep looping and paying until there are no remaining unpaid tokens.

packages/coil-extension/src/background/services/Stream.ts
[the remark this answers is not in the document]
sentientwaffle: You're right; my mistake. I'll add this.

packages/coil-anonymous-tokens/test/jest/redeemer.test.ts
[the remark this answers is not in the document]
sentientwaffle: I see, I misunderstood your initial comment. I'll rename the variables.

packages/coil-extension/src/background/services/Stream.ts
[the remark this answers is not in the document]
sublimator: Ah, I see, yeah, it's just doing 1 at a time, but the the top while loop condition will keep it continuing ?

sublimator:
@wilsonianb @sentientwaffle 

@sidvishnoi 
image

Brought up an interesting point in #webmonetization on the Interledger slack 

If we are relying on client side js to submit receipts to keep a balance, 
there may be some cases where  ... ?

wilsonianb:
I'm guessing a website could similarly never know about the final payment when it is one of smaller, more frequent payments.
The difference here is that the missed amount would be larger.
[Codius](https://github.com/codius/codius-faas) would be an example where the website tracks payments across sessions: A function's `remaining invocations` balance is credited based on submitted receipts for wm payments on the function's dashboard page.
I'm not sure what can be done to avoid missing the amount in that edge case.

sublimator:
> The difference here is that the missed amount would be larger.
Ja

packages/coil-extension/src/background/services/BackgroundScript.ts
sublimator: I think we should send both to be consistent

sentientwaffle:
For the oauth-scripts this isn't an issue, since it will pre-pay, not post-pay.

But the extension must post-pay to prevent sites from resetting the timer to get extra money. 

> If we are relying on client side js to submit receipts to keep a balance

Correct, they might miss the last receipt. It will never be more than 5 minutes of payment.

packages/coil-extension/src/background/services/Stream.ts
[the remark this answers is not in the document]
sentientwaffle: Yes.

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -769,10 +769,11 @@ export class BackgroundScript {
     const id = this.assoc.getStreamId(frame)
     if (id) {
       this.log('resuming stream', id)
-      this.sendSetMonetizationStateMessage(
-        frame,
-        this.hasStreamPaid(frame) ? 'started' : 'pending'
-      )
+      // Always set "pending" transition state and emit event
```
sentientwaffle, line 772: @sublimator Can you add an explanation (in the code) for why `pending` is always needed?

packages/web-monetization-wext/src/content/DocumentMonetization.ts
```diff
@@ -84,9 +84,15 @@ export class DocumentMonetization {
           this.window.location.origin
         )
       }
-      if (this.request && (state === 'stopped' || state === 'pending')) {
```
sentientwaffle, line 87: Should these changes be applied to oauth-scripts as well?

packages/web-monetization-wext/src/content/DocumentMonetization.ts
```diff
@@ -99,15 +105,10 @@ export class DocumentMonetization {
   }
 
   postMonetizationStartWindowMessageAndSetMonetizationState(
```
sentientwaffle, line 107: Should we just remove this method and have its callers call `setState` directly?

sublimator: Yeah, could/should inline it eh

I'll do it, but feel free, it's your PR! :)

packages/web-monetization-wext/src/content/DocumentMonetization.ts
sublimator: I'm not entirely sure about those exact changes, and the implications of pre vs post pay, but I believe they should both send pending/started in quick succession as per the extension

sublimator:
For github record
```
There seems to be a bug where 'started' events will be fired while the page is in a 'stopped' state in the background.
Load a page then quickly switch tabs before the start event is fired. You'll see a stop event then a few seconds after see a start event. I'm not sure what the answer is because it's post paying by design ofc.

Hrmmm, well, I mean the Loop has the wind down Ending state which it's handling all the time to postpay. It should really freeze all activity upon pause.

Needs pondering/fixing in any case
```

packages/coil-extension/src/background/services/BackgroundScript.ts
sublimator: I don't think it's *needed* per se (it's probably not) but who knows
This maintains the prior transitions, of which stopped->started wasn't part of

sublimator:
If you go to a page, quickly bind a `monetizationprogress` event handler, then quickly switch tabs before the start event is fired, you'll still get multiple progress events:
image

packages/coil-extension/src/content/services/ContentScript.ts
```diff
@@ -134,6 +135,11 @@ export class ContentScript {
         } else if (request.command === 'setMonetizationState') {
           this.monetization.setState(request.data)
         } else if (request.command === 'monetizationProgress') {
+          if (this.monetization.getState() === 'pending') {
+            // Don't transition from stopped → started. This prevents catch-up
```
sublimator, line 139: @sentientwaffle 
What's happening here ?
At a shallow reading, it seems implied that this supports progress events which may potentially happen when the state is stopped (i.e. tab isn't active) ?

sublimator: If you're deactivating a tab, should it not just pause everything ? It seems weird that it would suddenly "catchup" pay as soon as you switch the tab, and that the events should be emitted when the user is not watching (tab is not active). Especially when you consider it would otherwise wait?

Am I misunderstanding something

sentientwaffle: > At a shallow reading, it seems implied that this supports progress events which may potentially happen when the state is stopped (i.e. tab isn't active) ?

Yes, that may happen in order to prevent incorrectly activating a deactivated tab when a late payment arrived. And I figured emitting the progress event (even when state=stopped) was better than just discarding it (and much simpler than queuing them up until the resume).

> If you're deactivating a tab, should it not just pause everything ? It seems weird that it would suddenly "catchup" pay as soon as you switch the tab, and that the events should be emitted when the user is not watching (tab is not active). Especially when you consider it would otherwise wait?

Yeah, I'll fix that.

sentientwaffle:
@sublimator 

#### Batch info
- The very first payment (in both the extension and oauth-scripts) is paid gradually -- 5s, 5s, 20s, 30s (see: `firstMinuteBandwidth`).
  - This ensures that if the user is on a page for only a short time (e.g. before navigating away or closing a tab) a correspondingly small amount is paid.
- The batch/timer logic is in `PaymentScheduler.ts`.
- Time while the tab is activated is tracked. Initially `PaymentScheduler` will pay one token per minute, when enough time on the page has accumulated. After 10 minutes, it pays 2 tokens every 2 minutes (and so on, to a maximum of 5tokens/5mins).
- Extension pays at the end of an interval, oauth-scripts pays for a period of time beforehand.

#### Extension manual tests
- Closing a tab: When a tab is closed, the extension should send a "catch-up" payment of one or more tokens (the last one potentially being partial).
- Deactivating a tab: When a tab is deactivated, the extension doesn't send a catch-up payment. If after being deactivated, the tab is:
  - closed: A catch-up payment is sent for any unpaid time.
  - activated: The tab will resume its timer, and send payment when it has accumulated enough active time to merit a batch payment.
- Quickly switching through different tabs: afaik this is just a special case of "Deactivating a tab". Payments are only sent when batches are ready.
- Flicking back and forth between tabs: (see above)

#### Oauth-scripts manual tests:
- Closing a tab: n/a; the oauth-script can't run in a closed tab. Since oauth-scripts prepays, this is not an issue.
- Deactivating a tab: When a tab is deactivated, the extension doesn't send a catch-up payment. If after being deactivated, the tab is:
  - closed: (see above)
  - activated: The tab will start consuming its "prepaid time", and eventually pay another batch when its current prepaid time has run out. The "prepaid time" is not consumed while the tab is deactivated.

sublimator:
@sentientwaffle 

Very helpful! Thanks :)

sublimator:
```
(+0.000) Using Extension +0
(+0.005) event: window.load, initial state: stopped +5
(+0.005) event: window.load, current state: stopped +0
(+0.013) event: pending, state: pending requestId: 94bed7d0-8e36-4144-a0b9-fb95b4ef6635 pp=$twitter.xrptipbot.com/Coil +8
(+1.006) state: pending +993
(+2.009) state: pending +1003
(+3.010) state: pending +1001
(+4.011) state: pending +1001
(+4.774) event: start, state: started requestId: 94bed7d0-8e36-4144-a0b9-fb95b4ef6635 pp=$twitter.xrptipbot.com/Coil +763
(+4.775) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +1
(+5.008) state: started +233
(+5.078) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +70
(+5.389) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +311
(+5.692) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +303
(+6.006) state: started +314
(+6.056) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +50
(+6.409) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +353
(+6.717) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +308
(+7.009) state: started +292
(+7.043) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +34
(+7.428) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +385
(+7.733) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +305
(+8.006) state: started +273
(+8.254) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +248
(+8.560) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +306
(+8.849) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +289
(+9.007) state: started +158
(+9.131) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +124
(+9.424) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +293
(+9.789) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +365
(+10.008) state: started +219
(+10.075) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +67
(+10.366) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +291
(+10.710) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +344
(+11.006) state: started +296
(+11.031) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +25
(+11.325) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +294
(+11.632) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +307
(+11.940) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +308
(+12.007) state: started +67
(+12.231) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +224
(+12.554) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +323
(+12.861) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +307
(+13.009) state: started +148
(+13.169) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +160
(+13.475) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +306
(+13.782) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +307
(+14.011) state: started +229
(+14.091) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +80
(+14.397) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +306
(+14.704) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +307
(+15.007) state: started +303
(+15.011) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +4
(+15.303) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +292
(+15.626) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +323
(+15.932) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +306
(+16.006) state: started +74
(+16.246) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +240
(+16.547) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +301
(+16.852) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +305
(+17.011) state: started +159
(+17.164) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +153
(+17.576) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +412
(+17.891) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +315
(+18.006) state: started +115
(+18.188) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +182
(+18.491) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +303
(+18.800) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +309
(+19.006) state: started +206
(+19.078) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +72
(+19.376) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +298
(+19.722) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +346
(+20.010) state: started +288
(+20.032) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +22
(+20.336) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +304
(+20.644) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +308
(+20.968) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +324
(+21.011) state: started +43
(+21.258) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +247
(+21.567) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +309
(+21.869) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +302
(+22.011) state: started +142
(+22.177) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +166
(+22.483) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +306
(+22.792) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +309
(+23.006) state: started +214
(+23.099) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +93
(+23.406) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +307
(+23.716) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +310
(+24.010) state: started +294
(+24.023) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +13
(+24.293) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +270
(+24.637) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +344
(+24.945) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +308
(+25.007) state: started +62
(+25.251) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +244
(+25.559) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +308
(+25.855) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +296
(+26.006) state: started +151
(+26.172) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +166
(+26.559) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +387
(+26.860) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +301
(+27.006) state: started +146
(+27.187) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +181
(+27.471) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +284
(+27.812) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +341
(+28.007) state: started +195
(+28.118) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +111
(+28.428) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +310
(+28.702) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +274
(+29.007) state: started +305
(+29.039) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +32
(+29.341) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +302
(+29.654) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +313
(+29.962) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +308
(+30.008) state: started +46
(+30.236) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +228
(+30.576) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +340
(+30.873) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +297
(+31.008) state: started +135
(+31.191) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +183
(+31.500) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +309
(+31.829) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +329
(+32.008) state: started +179
(+32.318) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +310
(+32.668) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +350
(+32.955) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +287
(+33.009) state: started +54
(+33.240) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +231
(+33.549) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +309
(+33.852) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +303
(+34.011) state: started +159
(+34.161) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +150
(+34.466) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +305
(+34.775) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +309
(+35.007) state: started +232
(+35.090) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +83
(+35.356) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +266
(+35.696) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +340
(+35.985) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +289
(+36.008) state: started +23
(+36.311) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +303
(+36.605) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +294
(+36.930) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +325
(+37.007) state: started +77
(+37.232) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +225
(+37.540) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +308
(+38.011) state: started +471
(+38.054) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +43
(+38.358) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +304
(+38.668) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +310
(+38.973) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +305
(+39.008) state: started +35
(+39.280) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +272
(+39.569) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +289
(+39.895) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +326
(+40.006) state: started +111
(+40.192) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +186
(+40.508) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +316
(+40.821) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +313
(+41.007) state: started +186
(+41.110) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +103
(+41.431) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +321
(+41.738) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +307
(+42.007) state: started +269
(+42.059) event: progress, state: started amount: 5 code: USD scale: 9 receipt: undefined +52
(+43.007) state: started +948
(+44.007) state: started +1000
(+45.008) state: started +1001
(+46.007) state: started +999
(+47.008) state: started +1001
(+48.007) state: started +999
(+49.007) state: started +1000
(+50.008) state: started +1001
(+51.008) state: started +1000
(+52.007) state: started +999
(+53.007) state: started +1000
(+54.009) state: started +1002
(+55.008) state: started +999
(+56.008) state: started +1000
(+57.008) state: started +1000
(+58.008) state: started +1000
(+59.008) state: started +1000
(+60.009) state: started +1001
(+61.008) state: started +999
(+62.007) state: started +999
(+63.007) state: started +1000
(+64.008) state: started +1001
(+65.007) state: started +999
(+66.007) state: started +1000
(+67.010) state: started +1003
(+68.007) state: started +997
```

I added a delta from start for each line for clarity. 
It seems at least when I was throttled that the first minute was quite busy.

packages/coil-oauth-scripts/src/Stream.ts
```diff
@@ -316,7 +268,62 @@ export class Stream {
...
+
+function btpTokenThroughput(token: string): number {
+  const payload = Buffer.from(token.split('.')[1], 'base64')
+  const { throughput } = JSON.parse(payload.toString())
+  if (throughput === undefined) throw new Error('throughput not found')
+  return throughput
+}
+
+async function firstMinuteBandwidth(
+  stream: IlpStream.DataAndMoneyStream,
```
sublimator, line 295: Two things to note:
1. If the `stream` closes early for some reason (expired token reconnect?) this first minute behavior is skipped 
2. If it's started with a partially used token, won't that mean it won't get a full minutes worth?

sublimator: I'm not entirely sure the reason for it, but I've been inspecting the amounts given and more often than not, I don't see the final 30 seconds worth getting paid: https://gist.github.com/sublimator/88ecf832ee4454577108ecc27f1a5d0a#file-gistfile1-txt-L76

sublimator: This also means the first minute logic is interruptible. 
Open logger page, wait first payment event, flip to another tab, then back:
https://gist.github.com/sublimator/d62010f79c153e663a1018de568a329e#file-gistfile1-txt-L10

sublimator: And actually, it seems if you do that, it won't pay for minutes:
https://gist.github.com/sublimator/9bef500e6b3335c83ae25cff21deb372

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -116,133 +111,152 @@ export class Stream extends EventEmitter {
...
+          const waitTime = new Promise(
+            (_, reject) =>
+              (timer = setTimeout(() => reject(new Error('timeout')), 30e3))
+          )
+          // On error, destroy() was called internally by ilp-stream.
+          // Resolving here isn't necessary, but it prevents an extraneous destroy() call.
+          const waitFail = new Promise(resolve =>
+            connection.once('error', resolve)
+          )
+          await Promise.race([waitFail, waitTime, connection.end()])
```
sublimator, line 221: connection.end() waits for the stream.sendMax() ?

sentientwaffle: yes

packages/coil-extension/src/background/services/ActiveTabLogger.ts
```diff
@@ -28,4 +31,18 @@ export class ActiveTabLogger {
       }
     })
   }
+
+  sendLogEvent(message: string) {
```
sentientwaffle, line 35: What is the point of all of this? The background page has its own log.

sublimator:
Sadly, the code still seems buggy so I wanted to add some logging where I
could see it all easily in one place, timestamped and perhaps try to debug
it. Given I had already deployed to staging etc, it would have felt a
waste of time not to.

I actually didn't mean to push it to this branch. Sorry for that.

packages/coil-oauth-scripts/src/Stream.ts
sentientwaffle: > If the stream closes early for some reason (expired token reconnect?) this first minute behavior is skipped
> If it's started with a partially used token, won't that mean it won't get a full minutes worth?
> This also means the first minute logic is interruptible.

Correct. In each of those cases, it just means it falls back to post-pay a little early, which is fine.

> And actually, it seems if you do that, it won't pay for minutes
That's a bug; I'll push a fix.

sublimator: > it just means it falls back to post-pay a little early, which is fine

If it's "fine" do we even need this beyond the first payment ?

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -116,133 +113,158 @@ export class Stream extends EventEmitter {
...
-  async start() {
-    if (this._active) return
-    this._active = true
-
-    // reset this upon every start *before* early exit while _looping
+  async start(): Promise<void> {
+    // reset this upon every start *before* early exit while looping
     this._packetNumber = 0
+    this.isPaused = false
+    await this.loop.run(async (tokenFraction: number): Promise<Connection> => {
```
sublimator, line 141: @sentientwaffle 
It seems for some reason that this attempt callback passed in to run(...) is getting executed twice in quick succession. I don't recall seeing that before your latest changes, but I may have missed it ?

sublimator: Looking closer, I think it's the winding down of the previous stream.
When you refresh the page, it seems like it's trying to wind down 

Question: does it try to post pay in this situation?

sublimator: It seems it does, and successfully (at least the times I've looked)

packages/coil-extension/src/background/services/ActiveTabLogger.ts
sublimator: will remove it, just an easy way of viewing everything in one tab

packages/coil-extension/src/background/services/Stream.ts
sublimator: It was somewhat surprising that it seems to be connecting again, but that's working as intended, as the connection is discarded in between batches. 

So when you refresh the tab, it will close the stream, which will in turn put the stream in Ending state, and pay out whatever is necessary. This will (may?) require creating a new connection.

packages/coil-extension/src/background/services/Stream.ts
```diff
@@ -444,9 +439,7 @@ async function firstMinuteBandwidth(
     )
     logger.sendLogEvent(() => `firstMinuteBandwidth:${id}: waited ${delay}`)
     if (stopped) return
-    // FULL_TOKEN_AMOUNT essentially means burn whatever is left ...
```
sublimator: Lol, that was my comment right

sublimator:
@sentientwaffle 
Sorry for letting this bitrot 
Still intended to steal some stuff from it

Issue #1359, opened 2020-12-03
Festive iconography lost on refresh


sublimator:
![image](https://user-images.githubusercontent.com/525211/100973260-d159de80-356c-11eb-9631-49fe83b4bfee.png)

The hat disappears on non-monetized sites upon refresh, but is restored upon switching tabs

sublimator:
Fixed with https://github.com/coilhq/web-monetization-projects/pull/1375

sublimator:
This was not fixed in all cases it seems

sublimator:
The following url really seems to surface the issue consistently:
https://mail.google.com/mail/u/1/#inbox

sublimator:
This fix still isn't quite right.
Now, after the xmas period is over, the Santa hat flickers
Probably refreshing chrome will fix it (the default icon is set only once upon extension startup, though likely needs to be done hourly)

PR #1397, opened 2020-12-10
feat(coil-extension): fire monetizationprogress event on tipping

### LINKED ISSUE #900: Tipping should trigger 'monetizationprogress' event?
Currently, leaving a tip issues a non-standard event `tip`. (Which as a side-note should probably have a vendor prefix since it's non-standard, like `coilTip` or `coiltip`. For an example how different browser vendors handle vendor prefixes for JS events, see [CSS3 Animation Events](https://www.sitepoint.com/css3-animation-javascript-event-handlers/).)

I think it's fine to have a separate event for tipping but I think a monetizationprogress event should be triggered also. For instance, if a page is unlocking a reward when I pay a certain amount of money, my tips should count towards that.

That does require that the receipts for the tip are somehow passed from the Coil tipping service back to the client I suppose?

Lmk if you think I'm way off here or - if you agree - how we might implement this.

sabinebertram:
Closes #900, extension side of https://github.com/coilhq/coil/pull/4490

- [x] Be sure to send a progress event per receipt returned via franklin
- [x] Fix @coil/client modifications
- [x] Deal with modified polyfill ( "tip" event is removed )

### DIFF
--- a/packages/coil-client/src/graphQlClient.ts
+++ b/packages/coil-client/src/graphQlClient.ts
@@ -71,11 +71,15 @@ export class GraphQlClient {
       this.config.log('Domain:', this.config.coilDomain, 'Url:', redacted)
     }
     const res = await this.fetch(`${this.config.coilDomain}/graphql`, init)
-    if (!res.ok) {
+    // TODO: handle case when json fails
+    const gqlResponse = (await res.json()) as GraphQlResponse<T>
+    if (gqlResponse.errors?.length) {
       throw new Error(
-        `graphql query failed. status=${res.status} query=\`${query}\``
+        `graphql query failed. status=${
+          res.status
+        } query=\`${query}\` errors=${JSON.stringify(gqlResponse.errors)}`
       )
     }
-    return (await res.json()) as GraphQlResponse<T>
+    return gqlResponse
   }
 }
--- a/packages/coil-client/src/graphQlClient.ts
+++ b/packages/coil-client/src/graphQlClient.ts
@@ -10,7 +10,6 @@ import {
   whoAmI
 } from './queries'
 
-// Reference class for DI/reduct
 @injectable()
 export class GraphQlClientOptions {
   public coilDomain = 'https://coil.com'
@@ -22,12 +21,13 @@ export class GraphQlClientOptions {
 export interface GraphQlQueryParameters {
   query: string
   token?: string | null
+  autoThrow?: boolean
   variables?: Record<string, unknown>
 }
 
 @injectable()
 export class GraphQlClient {
-  public static Options = GraphQlClientOptions
+  public static readonly Options = GraphQlClientOptions
   protected readonly fetch: typeof fetch
 
   public login = login
@@ -38,7 +38,7 @@ export class GraphQlClient {
 
   public constructor(
     @inject(GraphQlClientOptions)
-    private config: GraphQlClientOptions = new GraphQlClientOptions()
+    protected config: GraphQlClientOptions = new GraphQlClientOptions()
   ) {
     this.fetch = this.config.fetch
   }
@@ -47,6 +47,7 @@ export class GraphQlClient {
   public async query<T = any>({
     query,
     token = null,
+    autoThrow = true,
     variables = {}
   }: GraphQlQueryParameters): Promise<GraphQlResponse<T>> {
     const init: RequestInit = {
@@ -71,15 +72,12 @@ export class GraphQlClient {
       this.config.log('Domain:', this.config.coilDomain, 'Url:', redacted)
     }
     const res = await this.fetch(`${this.config.coilDomain}/graphql`, init)
-    // TODO: handle case when json fails
-    const gqlResponse = (await res.json()) as GraphQlResponse<T>
-    if (gqlResponse.errors?.length) {
+    if (!res.ok && autoThrow) {
       throw new Error(
-        `graphql query failed. status=${
-          res.status
-        } query=\`${query}\` errors=${JSON.stringify(gqlResponse.errors)}`
+        `graphql query failed. status=${res.status} ` +
+          `query=\`${query}\` text=${JSON.stringify(await res.text())}`
       )
     }
-    return gqlResponse
+    return (await res.json()) as GraphQlResponse<T>
   }
 }
--- a/packages/coil-extension/src/background/services/BackgroundScript.ts
+++ b/packages/coil-extension/src/background/services/BackgroundScript.ts
@@ -48,6 +48,16 @@ import { debug } from '../../content/util/logging'
 
 import { ActiveTabLogger } from './ActiveTabLogger'
 
+interface SendTipResult {
+  sendTip: {
+    success: boolean
+    amount: string
+    currency: string
+    scale: number
+    receipts: string[]
+  }
+}
+
 @injectable()
 export class BackgroundScript {
   constructor(
@@ -727,10 +737,10 @@ export class BackgroundScript {
 
     try {
       this.log(`sendTip: sending tip to ${receiver}`)
-      const result = await this.client.query({
+      const result = await this.client.query<SendTipResult>({
         query: `
-          mutation sendTip($receiver: String!) {
-            sendTip(receiver: $receiver, requestId: $streamId) {
+          mutation sendTip($receiver: String!, $amount: Int, $requestId: String) {
+            sendTip(receiver: $receiver, amount: $amount, requestId: $requestId) {
               success
               amount
               currency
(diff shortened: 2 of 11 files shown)


packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -728,6 +728,7 @@ export class BackgroundScript {
         command: 'tip',
         data: {
           paymentPointer: receiver,
+          requestId: streamId,
```
sublimator, line 731: This implies the SPSP server not really doing anything relying on the Web-Monetization-Id in terms of accounting ?
It's already been deprecated right

tsconfig.json
```diff
@@ -1,50 +1,27 @@
...
-      "@coil/privacypass-sjcl": [
-        "packages/coil-privacypass-sjcl/src"
-      ],
-      "@coil/webpack-utils": [
-        "packages/coil-webpack-utils/src"
-      ],
-      "@dier-makr/annotations": [
-        "packages/dier-makr-annotations/src"
-      ],
+      "@coil/monorepo-upkeep": ["packages/coil-monorepo-upkeep/src"],
```
sublimator, line 8: run `yarn upkeep`, it should format these in the canonical format expected by CI

packages/coil-extension/src/background/services/BackgroundScript.ts
sublimator: Should the requestId be sent to the Coil backend to send as Web-Monetization-Id to the SPSP server when sending the tip? For backend integrations that aren't yet using receipts ?

packages/coil-extension/src/content/services/ContentScript.ts
```diff
@@ -161,13 +161,14 @@ export class ContentScript {
           this.onFrameAllowedChanged(request)
         } else if (request.command === 'tip') {
           debug('sendTip event')
-          const detail: TipEvent['detail'] = {
+          const detail: MonetizationProgressEvent['detail'] = {
             amount: request.data.amount,
             assetCode: request.data.assetCode,
             assetScale: request.data.assetScale,
-            paymentPointer: request.data.paymentPointer
+            paymentPointer: request.data.paymentPointer,
```
sublimator, line 168: Backend not yet sending receipts ?

tsconfig.json
sublimator: pushed !

sabinebertram: Thanks! I was wondering what I need to do to fix this issue.

packages/coil-extension/src/content/services/ContentScript.ts
sabinebertram: We are not passing them through yet. I'm tackling that next.

packages/coil-extension/src/background/services/BackgroundScript.ts
sabinebertram: I'm afraid I don't understand what you are asking.

sublimator: Sorry, bad sleep! I'm going to go with that :)

Depending on the WM integration, it may be that the backend tracks balances via the requestId sent to the SPSP server as "Web-Monetization-Id", so this means that in some cases, checking the balance on the backend by requestId won't show the tip. Sigh, I realize that's a bit vague, but that's because we're talking about potential integrations we can't really point to and discuss?

sharafian: It sounds like the potential issue here in most cases we have `Web-Monetization-Id` sent on the SPSP request when we make payments, and then it's emitted on the monetizationprogress event so if the server wants to correlate SPSP request/payment to the monetizationprogress event, it can.

However, Franklin doesn't send the `Web-Monetization-Id` header, so emitting a requestId on this payment may be a bit disingenuous. The site may try to cross reference that requestId with a payment only to find that no SPSP requests were made with the `Web-Monetization-Id` header.

But with the `Web-Monetization-Id` feature being deprecated (in favor of submitting receipts to verify payment) we could possibly include the requestId to avoid breaking the interface without actually sending it, with the assumption nobody will check that `Web-Monetization-Id` was really sent.

It might be better, though, just to be thorough and actually allow the `Web-Monetization-Id`/requestId to be passed to the API, then to Franklin, and then include it in Franklin's SPSP request. I don't think it would be very hard to add as an optional field and it's one less thing to worry about.

sublimator: Ok, I will hold off on merging then

sublimator (changes requested):
Holding until backend graphql mutation returns receipts and takes an optional requestId

packages/coil-extension/src/background/services/BackgroundScript.ts
sabinebertram: > we could possibly include the requestId to avoid breaking the interface without actually sending it

Include it in the `monetizationprogress` event without sending it via Franklin in the SPSP request? Just making sure I get it.

> It might be better, though, just to be thorough and actually allow the Web-Monetization-Id/requestId to be passed to the API, then to Franklin, and then include it in Franklin's SPSP request.

It sounds cleaner to me.

sublimator: > Include it in the monetizationprogress event without sending it via Franklin

That's basically the state of the PR now (if I'm not confusing myself)

sublimator: > Ok, I will hold off on merging then

@sharafian 
With an eye ((perhaps foolishly optimistic)) to getting a release out before the year's end?
With the fix for #1359 (and in that vein, some fireworks 31/12) and maybe #1167 now that @fruehle is back from holidays

sabinebertram:
Depends on coil api changes to be merged ([#4490](https://github.com/coilhq/coil/pull/4490)).

packages/coil-extension/src/content/services/ContentScript.ts
```diff
@@ -161,13 +161,15 @@ export class ContentScript {
...
+            requestId: request.data.requestId,
+            receipt: request.data.receipts ? request.data.receipts[0] : ''
```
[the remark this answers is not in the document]
sublimator: @wilsonianb 
For some reason I'm remembering you as the person who altered Franklin to return receipts?

sublimator (changes requested):

packages/coil-extension/src/content/services/ContentScript.ts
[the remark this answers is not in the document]
wilsonianb: Yeah, I think the intention was for there to be a `monetizationprogress` emitted for each returned receipt
https://github.com/coilhq/coil/pull/4304#issuecomment-722018581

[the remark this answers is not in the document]
sublimator: @wilsonianb 

Great, thanks for the confirm/info

packages/web-monetization-wext/src/content/wmPolyfill.ts
```diff
@@ -1,7 +1,8 @@
 const events = [
   'monetizationstart',
   'monetizationstop',
-  'monetizationpending'
+  'monetizationpending',
```
sublimator: need a CI test that checks the manifest.json hash listed in the CSP matches the hashed polyfill

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -724,36 +734,50 @@ export class BackgroundScript {
     }
 
     const receiver = stream.getPaymentPointer()
-    const { assetCode, assetScale, exchangeRate } = stream.getAssetDetails()
```
sublimator, line 727: See: https://github.com/coilhq/web-monetization-projects/pull/567#discussion_r419924429

Removed as the new tipping endpoint will return the exact amount sent

packages/coil-extension/src/content/services/ContentScript.ts
```diff
@@ -160,14 +160,26 @@ export class ContentScript {
         } else if (request.command === 'onFrameAllowedChanged') {
           this.onFrameAllowedChanged(request)
         } else if (request.command === 'tip') {
-          debug('sendTip event')
-          const detail: TipEvent['detail'] = {
+          debug('sendTip event', request.data)
+          const detail: MonetizationProgressEvent['detail'] = {
```
sublimator, line 164: image

This doesn't seem right

sublimator: https://github.com/coilhq/coil/pull/4490/files#r589991430

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -724,36 +734,63 @@ export class BackgroundScript {
...
+              }
             }
           }
         `,
         token,
         variables: {
-          receiver
+          receiver,
+          amount: undefined, // default
+          webMonetizationId: streamId
```
wilsonianb, line 761: Would `streamId` be the same `requestId` used for web monetization's `Web-Monetization-Id`?

packages/coil-client/src/graphQlClient.ts
```diff
@@ -38,7 +38,7 @@ export class GraphQlClient {
 
   public constructor(
     @inject(GraphQlClientOptions)
-    private config: GraphQlClientOptions = new GraphQlClientOptions()
+    protected config: GraphQlClientOptions = new GraphQlClientOptions()
```
wilsonianb, line 41: Is this change needed?

wilsonianb (approved):
LGTM :+1:

packages/coil-client/src/graphQlClient.ts
sublimator: no, not needed, but it's not "incorrect"
it was cause I'd modified these files previously in another branch
so I just pinched the changes
I'm not religious about confining small changes to their own commits

packages/coil-extension/src/background/services/BackgroundScript.ts
sublimator: Yep, streamId is what will be used for WMID and the requestId in the events

sublimator:
Closing, as Bernie won't support progress events due to async sending of funds:
https://github.com/coilhq/coil/pull/4714#issuecomment-828221546

PR #1503, opened 2021-01-22
feat(ext): polyfill via document events rather than posting window messages

sublimator:
CI *was* failing due to the `detail` property being restricted on CustomEvents on Firefox 
image

There seems to be potential workarounds listed here:
https://stackoverflow.com/questions/18744224/triggering-a-custom-event-with-attributes-from-a-firefox-extension

Simply sending detail as a string (JSON serialized object) seems to be simpler than using the cloneInto mentioned on SO

Using cloneInto regardless as it's less "work" done in the content script thread

### DIFF
--- a/packages/web-monetization-wext/src/content/DocumentMonetization.ts
+++ b/packages/web-monetization-wext/src/content/DocumentMonetization.ts
@@ -27,6 +27,13 @@ type DefaultView = WindowProxy & typeof globalThis
 type CloneInto = (obj: unknown, window: DefaultView | null) => typeof obj
 declare const cloneInto: CloneInto | undefined
 
+let cloneIntoRef: CloneInto | undefined
+try {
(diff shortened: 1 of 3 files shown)


packages/web-monetization-wext/src/content/DocumentMonetization.ts
```diff
@@ -20,17 +20,27 @@ interface SetStateParams {
...
+
+type DefaultView = WindowProxy & typeof globalThis
+type CloneInto = (obj: unknown, window: DefaultView | null) => typeof obj
+declare const cloneInto: CloneInto | undefined
+
+let cloneIntoRef: CloneInto | undefined
+try {
+  cloneIntoRef = cloneInto
+} catch (e) {
+  cloneIntoRef = undefined
```
sublimator, line 34: Will throw on chrome if try to access missing variable

PR #1641, opened 2021-02-28
feat(coil-webpack-utils): add getPackageVersion

sublimator:
- [x] add date to Version struct


packages/coil-oauth-scripts/src/MonetizationPolyfill.ts
```diff
@@ -57,7 +67,7 @@ export class MonetizationPolyfill {
     this.btpEndpoint = opts.btpEndpoint
 
     const streamControl: StreamControl = {
-      pause: () => this.stream.pause(),
+      pause: async () => this.stream.pause(),
```
sublimator, line 70: hrmmm, an eslint autofix

PR #1654, opened 2021-03-02
feat: quick hack to distribute tips

### LINKED ISSUE #1414: Tipping when top level frame is not monetized, only iframe[s] are
> Once you run one of the games it fires the iframe monetization start and the extension lights up. However ever time I attempt to tip the game/content, to extensions says ’something went wrong.’… example games:
https://havana24.itch.io/relic-splatter
https://nimblebeastscollective.itch.io/moonshot

> ALSO: could we do something about that error message, ie. today it flashes in and out so it’s easy to miss. Perhaps if there is ‘something wrong’ it’s in a different colo(u)r and stays persistent so it’s more highlighted. And a way to easily report it — or to who.

sublimator:
Obviously not to be merged.
Something like this could have fixed #1414

### DIFF
--- a/packages/coil-extension/src/background/services/BackgroundScript.ts
+++ b/packages/coil-extension/src/background/services/BackgroundScript.ts
@@ -705,59 +705,78 @@ export class BackgroundScript {
 
   private async sendTip(): Promise<{ success: boolean }> {
     const tabId = this.activeTab
-    const streamId = this.assoc.getStreamId({ tabId, frameId: 0 })
-    if (!streamId) {
-      this.log('can not find top frame for tabId=%d', tabId)
-      return { success: false }
-    }
-    const stream = this.streams.getStream(streamId)
-    const token = this.auth.getStoredToken()
 
-    // TODO: return detailed errors
-    if (!stream || !token) {
-      this.log(
-        'sendTip: no stream | token. !!stream !!token ',
-        !!stream,
-        !!token
-      )
-      return { success: false }
-    }
+    const streams = Object.entries(this.assoc.getTabStreams(tabId))
+    const numTips = streams.length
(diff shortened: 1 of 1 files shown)


packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -705,59 +705,78 @@ export class BackgroundScript {
...
-        !!token
-      )
-      return { success: false }
-    }
+    const streams = Object.entries(this.assoc.getTabStreams(tabId))
+    const numTips = streams.length
+    const success: boolean[] = []
+    const messages: TipSent[] = []
+
+    for (const [frameIdStr, streamId] of streams) {
```
sublimator, line 714: If one was to do this properly, one should loop through the streams and perhaps wait until the exchange rate was set ( in the case of streams that are just activating - note that with iframes - they tend to connect in a staggering manner) 

Another thing to consider is that in the UI currently the tip button is only shown  if the stream has activated (so the exchange rate is shown). Would need to consider multiple frames.

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -705,59 +705,78 @@ export class BackgroundScript {
...
+        this.log(`sendTip: error. msg=${e.message}`)
+        // return { success: false }
       }
-      this.api.tabs.sendMessage(tabId, message, { frameId: 0 })
-      return { success: true }
-    } catch (e) {
-      this.log(`sendTip: error. msg=${e.message}`)
-      return { success: false }
     }
+    // As long as one tip went through? success
```
sublimator, line 775: > // As long as one tip went through? success

Just for hack

PR #1659, opened 2021-03-04
feat(coil-extension): impl sec-gpc: 1 via declarativeNetRequest

sublimator:
Investigate using declarativeNetRequest for https://globalprivacycontrol.org/ 
- [x] send sec-gpc header 
- [x] ability to toggle manifest declared rule programmatically
  - https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#method-updateEnabledRulesets
- [ ] document the permissions request ux

### DIFF
--- a/packages/coil-extension/gpc.json
+++ b/packages/coil-extension/gpc.json
@@ -1,0 +1,21 @@
+[{
+  "id": 1,
+  "priority": 1,
+  "action": {
+    "type": "modifyHeaders",
+    "requestHeaders": [
+      {
+        "header": "Sec-GPC",
+        "operation": "set",
+        "value": "1"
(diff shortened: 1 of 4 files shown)


adrianhopebailie:
I assume this is to allow us to provide an option for users to send the https://globalprivacycontrol.org/ header?

sublimator:
@adrianhopebailie 
Yep, was just looking at what it took
Last time I took a quick look at it I used declarativeWebRequest rather than declarativeNetRequest and I couldn't get it to work
It's a good point that it should be optional
When I get a chance, I'll see about programmatically toggling the rule (I assume you can reference the declared rule id)

PR #1717, opened 2021-03-17
feat(yarn): update yarn to 2.4.1

sublimator:


sublimator:
So once upon a time we configured .gitignore to ignore .pnp.js, but then yarn started calling it .pnp.cjs and it was added to the repository. It's once again ignored now that 2.4.1 has restored the original name.

PR #1718, opened 2021-03-17
feat(vscode): add pnpify --sdk vscode

sublimator:


.yarn/sdks/typescript/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "typescript",
+  "version": "4.2.3-pnpify",
```
sublimator, line 3: I wonder if we should have some step in the CI to keep the version up to date, so it doesn't eventually get more and more misleading

PR #1340, opened 2020-11-26
feat(coil-extension): safari workaround for missing webNavigation apis etc

sublimator:
TODO:
- [x] workarounds for missing apis as in #1076
- [x] use window 'storage' event rather than localStorageUpdate sent via browser.runtime.sendMessage
   - Solve the `localStorage==null` issue #1077  
- [x] some background -> popup command mechanism
  - maybe window dom events ?
     - ~~send via chrome.extension.getViews({type: 'popup'}) from the bg page ?~~
       - does not work
  - [x] localStorage based ? serialized command and react using window 'storage' event ?
  - window.close() in response to {command: 'closePopup} from bg also causes a bad state! (#1077)
    - [x] Find a way to detect safari browser
- [x] refactor to use some kind of contextual event bus 
   -  should support multiple instances of the popup in one window
- [x] Open issue on Apple Feedback Assistant re: popup communication woes
  - https://feedbackassistant.apple.com/feedback/8926137
- [x] Open issue re: missing apis
   - https://feedbackassistant.apple.com/feedback/8730949

PR #1786, opened 2021-04-08
feat(coil-extension): refactor popup

sublimator:
- Refactor popup which passed down a context prop everywhere to use React context providers/consumers. 
- Split PopupContext into PopupHost and Store contexts. 
- Add "events" (EventEmitter) to PopupHost to support multiple instances of a popup in a page (styling aid)
- Remove extension message listeners for `localStorageUpdate` messages (*to support Safari*, as per #1340 )
  - Listen to "storage" event on new "events" bus in PopupHost
     - Extension Host: bind to window "storage" event then proxy to `events` bus
     - Mock Host: emit mock "storage" events directly to trigger payment animation with updated monetizationTotal
- Add useStoreState hook to create Store to pass as prop to StoreContext.Provider
  - Parameterized with get only (JSON parsing) Storage wrapper and PopupHost 
     - listens for "storage" events, updating store accordingly

### DIFF
--- a/packages/coil-extension/src/popup/components/MonetizationAnimation.tsx
+++ b/packages/coil-extension/src/popup/components/MonetizationAnimation.tsx
@@ -1,12 +1,15 @@
 import React, { useEffect, useRef, useState } from 'react'
 
-import { PopupProps } from '../types'
-import { ToPopupMessage } from '../../types/commands'
 import { notNullOrUndef } from '../../util/nullables'
+import { StorageEventPartial, useStore } from '../context/storeContext'
+import { useHost } from '../context/popupHostContext'
 
 const ANIMATION_INTERVAL = 1800
 
-export const MonetizeAnimation = (props: PopupProps) => {
+export const MonetizeAnimation = () => {
+  const store = useStore()
+  const host = useHost()
+
   const [lastPacket, setLastPacket] = useState<Date | null>(null)
   const [animated, setAnimated] = useState<boolean | null>(false)
   const lastPacketRef = useRef(lastPacket)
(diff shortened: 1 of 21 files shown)


packages/coil-extension/src/popup/components/Status.tsx
```diff
@@ -1,25 +1,21 @@
...
-      return <Unsubscribed context={context} />
+    if (user.subscription?.active) {
+      return <PaidViews />
     } else {
-      return <PaidViews context={context} />
+      return <Unsubscribed />
     }
   } else {
-    return <LoggedOut context={context} />
+    // Popup is never shown when logged out
```
sublimator, line 18: Is that actually true?

packages/coil-extension/src/popup/components/MonetizationAnimation.tsx
```diff
@@ -28,11 +31,12 @@ export const MonetizeAnimation = (props: PopupProps) => {
...
 
-    const listener = (msg: ToPopupMessage) => {
+    const listener = (msg: StorageEventPartial) => {
       if (
-        msg.command === 'localStorageUpdate' &&
+        msg.key &&
+        msg.newValue &&
         msg.key === 'monetizedTotal' &&
-        props.context.store.monetizedTotal > 0
+        JSON.parse(msg.newValue) > 0
```
sublimator, line 39: store would be stale at this point inside useEffect

packages/coil-extension/src/popup/components/StreamControls.tsx
```diff
@@ -152,37 +153,40 @@ interface SetStreamControlsParams {
...
-      if (message.command === 'localStorageUpdate') {
-        if (message.key === 'stickyState') {
-          const sticky = props.context.store.stickyState
-          // TODO: document why have and why need to ignore null changes
-          sticky != null && setStickyState(sticky)
-        } else if (message.key === 'playState') {
-          const play = props.context.store.playState
-          play != null && setPlayOrPauseState(play)
-        }
+    const listener = (message: StorageEventPartial) => {
```
sublimator, line 176: can make these nicer in a subsequent changeset

mwill8886 (approved):

PR #1808, opened 2021-04-14
feat: tipping UI

mwill8886:
- Tipping UI 
  - Refactored the components directory to include view specific files
  - Refactored view logic
  - Added tipping views
  - Added tipping UI animation
- Tipping functionality
  - Modified `sendTip` method to accept an amount
  - Stubbed in `setTipSettings` with some placeholder values

**Note**
I fetched latest from `main` and merged `main` into `mw-tipping`. It is now showing a bunch of the changes that should already be in main for all the `package.json` updates.


sublimator:
Looks like this is using old versions of the views:
image

Should be:
image

See the figma:
https://www.figma.com/file/QBDNeSLdmN52wreAJlgdiKOC/Extension?node-id=0%3A1

You should be able to see the relevant changes with:
`git diff 820955be62e565fdc29f148504efea883a94486d aaba672fe810c71a77ba2fd65ae08c9c50bbf76e src/popup`
the first commit is just before your branch started and the second commit is just before the latest popup refactor

You could build on `main` and then save the `dist/` to some other folder, then use the mock popups page as a reference. Then switch to your branch and just try and roll in those changes, after doing git merge origin/main -Xours (you will have a few stragglers, and some missing extension package.json entries to deal with then)

PR #1815, opened 2021-04-15
refactor: popup view components

mwill8886:
- Refactored popup view components
  - Created `views` directory for view components
  - Moved view components into `views` directory and changed names to include "View"
  - Moved view logic into view "routers" that live inside the `views` directory
  - Created `HeaderFooterLayout` component that gets consumed by view components


packages/coil-extension/src/popup/components/views/ViewRouter.tsx
```diff
@@ -0,0 +1,51 @@
...
+  //* Moved all the render logic from Status, PaidViews, CoilViews, and MonetizedPage here to Index
+  //* Not sure how I feel about the readability vs the original method.
+
+  //
+  // Invalid user views
+  //
+  if (!validToken && !user) {
+    return <LoggedOutView />
+  }
+  if (!user?.subscription || (user.subscription && !user.subscription.active)) {
```
sublimator, line 24: could do (!user?.subscription?.active)

sublimator:
@fruehle 

This seems OK at a glance to me, but could you check:
https://web-monetization-projects-git-mw-view-refactor-coilhq.vercel.app/static/popup.html

?

PR #1897, opened 2021-05-13
fix(coil-client): fix tokenUtils#newestToken

sublimator:


packages/coil-client/test/jest/tokenUtils.test.ts
```diff
@@ -68,6 +68,20 @@ describe('CoilTokenUtils', () => {
       expect(newest.which).toBe('site')
     })
 
+    it('should pick the token with the greatest expiry', () => {
+      const extensionToken: DecodedToken = { exp: 16, userId: 'custA', iat: 1 }
```
sublimator, line 72: no, it's specifically the "newest" token ...

PR #1910, opened 2021-05-14
feat(coil-extension): add accept-payment sketch

mankins:
This PR proposes the addition of the `Accept-Payment` header which follows the Content Negotiation pattern used by `Accept`, `Accept-Language`, etc.

If this PR is merged, the following header will be added to web requests from the client:

```
Accept-Payment: webmon/*;q=.8
```

HTTP servers can then choose specialized content for the client, or issue a content negotiation sequence: a 402 status code followed by user-facing instructions on how they might pay for the site.

More on the [Accept-Payment proposal](https://github.com/mankins/accept-payment).

This is opportunistic and doesn't replace the client-side web monetization standard.

This is a signal, not a guarantee. Well behaved clients can get the benefit of individual content without having to wait for detection / complete page load.

# Verify / QA
The see this in action you can build the extension and then inspect the outgoing http headers. There should be a new header:

```
Accept-Payment: webmon/*;q=.8
```

To test how content negotiation might work you can run the express server at [Accept-Payment proposal](https://github.com/mankins/accept-payment) and visit the pages referenced for web monetization.

PR #1916, opened 2021-05-16
feat(coil-extension): add accept-payment via declarativeNetRequest

mankins:
Note this a different approach to https://github.com/coilhq/web-monetization-projects/pull/1910, as suggested in the comments by @sublimator https://github.com/coilhq/web-monetization-projects/pull/1910#discussion_r633034791

This PR proposes the addition of the `Accept-Payment` header which follows the Content Negotiation pattern used by `Accept`, `Accept-Language`, etc.

If this PR is merged, the following header will be added to web requests from the client:

```
Accept-Payment: webmon/*;q=.8
```

HTTP servers can then choose specialized content for the client, or issue a content negotiation sequence: a 402 status code followed by user-facing instructions on how they might pay for the site.

More on the [Accept-Payment proposal](https://github.com/mankins/accept-payment).

This is opportunistic and doesn't replace the client-side web monetization standard.

This is a signal, not a guarantee. Well behaved clients can get the benefit of individual content without having to wait for detection / complete page load.

# Verify / QA
The see this in action you can build the extension and then inspect the outgoing http headers. There should be a new header:

```
Accept-Payment: webmon/*;q=0.8
```

To test how content negotiation might work you can run the express server at [Accept-Payment proposal](https://github.com/mankins/accept-payment) and visit the pages referenced for web monetization.

### DIFF
--- a/packages/coil-extension/manifest.json
+++ b/packages/coil-extension/manifest.json
@@ -23,6 +23,13 @@
   "web_accessible_resources": [
 
   ],
+  "declarative_net_request": {
+    "rule_resources": [{
+      "id": "ruleset_1_accept_payment",
+      "enabled": true,
+      "path": "static/rules.json"
+    }]
+  },
   "content_scripts": [
     {
       "matches": ["https://*/*","http://*/*"],
@@ -32,5 +39,5 @@
       "run_at": "document_start"
     }
   ],
-  "permissions": [ "webNavigation", "<all_urls>" ]
+  "permissions": [ "webNavigation", "declarativeNetRequest", "<all_urls>" ]
 }
--- a/packages/coil-extension/static/rules.json
+++ b/packages/coil-extension/static/rules.json
@@ -1,0 +1,13 @@
+[
+  {
+    "id": 1,
+    "priority": 1,
+    "action": {
+      "type": "modifyHeaders",
+      "requestHeaders": [
+        { "header": "accept-payment", "operation": "set", "value": "webmon/*;q=0.8" }
+      ]
+    },
+    "condition": { "resourceTypes": ["main_frame"] }
+  }
+]


packages/coil-extension/static/rules.json
```diff
@@ -0,0 +1,13 @@
+[
+  {
+    "id": 1,
+    "priority": 1,
+    "action": {
+      "type": "modifyHeaders",
+      "requestHeaders": [
+        { "header": "accept-payment", "operation": "set", "value": "webmon/*;q=0.8" }
```
mankins, line 8: Ideally this `set` would be an `append` but apparently there's an [open bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1117475) that prevents that

packages/coil-extension/manifest.json
```diff
@@ -23,6 +23,13 @@
   "web_accessible_resources": [
 
   ],
+  "declarative_net_request": {
+    "rule_resources": [{
+      "id": "ruleset_1_accept_payment",
+      "enabled": true,
+      "path": "static/rules.json"
```
mankins, line 30: Not sure if this is the right place for the file. I saw a different approach in another PR, but it wasn't merged.

sublimator: static/rules.json seems good to me

sublimator: well, I mean I guess you could more specific than "rules", but inside static is good

mankins: Updated `rules.json` -> `acceptPaymentRules.json`

packages/coil-extension/static/rules.json
sublimator: Will it matter much in practice do you think ? 
There will be races for which extension gets to set it in the case of more than one enabled ?

mankins: In the start it won't matter.

Eventually it could be a race condition. Ideally you'd want an adblock extension to be able to say "no ads" as well as what we have here. I'd think by the time those extensions implement and there's a supporting ux the bug would be fixed.

sublimator:
@mankins 

Hey, so of course Firefox doesn't support the declarativeNetRequest api/permission :/ 

It's a pity cause it's a decent API with performance advantages (one assumes at least)

mankins:
@sublimator I could add back https://github.com/coilhq/web-monetization-projects/pull/1910 into this PR?

I haven't looked into how the different builds happen, but you could branch if the browser has support? Or does this make the proposal DOA?

sublimator:
Yes, we could make that work! :)
I can look into hacking the build for you.
I just forgot before its not impl on FF. A pity Mozilla is so against it.

packages/coil-extension/static/acceptMonetizationRules.json
```diff
@@ -5,7 +5,7 @@
     "action": {
       "type": "modifyHeaders",
       "requestHeaders": [
-        { "header": "accept-payment", "operation": "set", "value": "webmon/*;q=0.8" }
+        { "header": "accept-monetization", "operation": "set", "value": "webmon/*;q=0.8" }
```
mankins, line 8: It's been suggested that "monetization" is a better name, so updating to align with other documents

sublimator: nice

sublimator:
@mankins 
I guess you saw this:
https://blog.mozilla.org/addons/2021/05/27/manifest-v3-update/

Mozilla are implement declarativeNetRequest :)

PR #1918, opened 2021-05-17
fix(coil-extension): refresh tokens eagerly

### LINKED ISSUE #1917: Refresh token periodically such that the token always has 27-28 days validity
If you don't visit coil.com or monetized sites frequently enough, you won't get a refreshed token. While the extension DOES refresh tokens, because it (foolishly, it seems) tries not to refresh too often, a short window is created, where if you don't trigger a refresh in that period (by visiting a monetized site) then the token will expire. The fix would be to more eagerly refresh, triggered on an interval, not just by visiting a monetized site.

sublimator:
Fixes #1917


packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -31,6 +70,21 @@ export class AuthService extends EventEmitter {
...
+  token is always valid for at least 26-28 days.
+   */
+  startTokenRefreshInterval() {
+    const twelveHours = 12 * 60 * 60 * 1e3
+    // Add some randomness to the interval so we
+    // can't correlate when a user is active/inactive quite as easily
+    const randomness = Math.random() * twelveHours
+    setInterval(() => {
+      void this.getTokenMaybeRefreshAndStoreState()
+    }, twelveHours + randomness)
```
sharafian, line 85: this isn't a very big deal, but it might be better to have a random delay after each refresh rather than calculating a random interval and maintaining it for the lifetime of the extension regularly

sharafian (changes requested):

packages/coil-extension/src/background/services/AuthService.ts
sublimator: yes, definitely

packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -10,6 +10,45 @@ import { SiteToken } from './SiteToken'
...
+ - The extension will listen to coil_writeToken events which can be emitted
+ with an empty string in the token field when logged out. In this case the
+ extension will clear its token too.
+
+ - If the extension sees an empty or null token on coil.com and the extension
+ has a token (because of logging out while the extension was disabled
+ and missing the logout event) it will inject the extensions token into the
+ site. This supports convenient use of incognito contexts.
+
+ We could do this only for incognito contexts and instead logout in normal
```
sublimator, line 42: @sharafian 
Should we go to some more lengths to try and handle the case of logging out when the extension is disabled? I guess upon startup of the extension we could use the SiteToken service to get the token and logout if there isn't any ? That seems KISS enough.

sublimator: "May as well implement that"
"Oh, but then when you're developing ... "

I'm inclined to leave it as is. On the odd occasion it does pop up, it's likely to be noticed and you can just log out again?

sublimator: Can always disable the check in DEV mode ?

sharafian: yeah that sounds reasonable

packages/coil-extension/src/background/services/SiteToken.ts
```diff
@@ -18,10 +18,34 @@ import { timeoutRejecting } from '../../util/timeout'
...
+    private coilDomain: string,
+    @inject(tokens.WextApi)
+    private api = chrome
   ) {}
 
   async retrieve(path = '/handler.html'): Promise<string | null> {
+    return new Promise(resolve => {
+      this.api.tabs.create(
+        {
+          active: false,
```
sublimator, line 30: @sharafian 

This actually isn't too bad UX. It opens a tab without switching your focus at all, then closes it once it has retrieved the token. If [restoring handler.html](https://github.com/coilhq/coil/pull/4865) doesn't resolve the issue on [Firefox](https://github.com/coilhq/web-monetization-projects/issues/1941#issuecomment-849288125) perhaps we could live with this?

packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -108,8 +108,24 @@ export class BackgroundScript {
...
-    // noinspection ES6MissingAwait
-    void this.auth.getTokenMaybeRefreshAndStoreState()
+    void this.initAuth()
+  }
+
+  private async initAuth() {
+    // API only available on FF
+    // https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBrowserInfo
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const isFF = Boolean((this.api.runtime as any)['getBrowserInfo']) || true
```
sublimator, line 118: TODO

packages/coil-extension/src/background/services/SiteToken.ts
sublimator: For the record, the handler fix fixed the freeze on FF

PR #1927, opened 2021-05-19
feat(coil-extension): support Safari via workarounds

sublimator:
Update of https://github.com/coilhq/web-monetization-projects/pull/1340

* Requires "tabs" permission:
   Chrome (and to a lesser extent, Firefox) grants many "tabs" permissions
   implicitly via "<all_urls>". Safari does not. This is required for using
   chrome.tabs to search for an already open tab via url and select it.
   
   via `export WEXT_MANIFEST_PERMISSIONS='["+tabs"]'`

### DIFF
--- a/packages/coil-extension/manifest.json
+++ b/packages/coil-extension/manifest.json
@@ -32,5 +32,10 @@
       "run_at": "document_start"
     }
   ],
-  "permissions": [ "webNavigation", "<all_urls>" ]
+  "permissions": [ "webNavigation", "<all_urls>" ],
+  "$targets" : {
+    "safari": {
+      "permissions" : ["tabs"]
+    }
+  }
 }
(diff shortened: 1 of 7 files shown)


packages/coil-extension/manifest.json
```diff
@@ -32,5 +32,5 @@
       "run_at": "document_start"
     }
   ],
-  "permissions": [ "webNavigation", "<all_urls>" ]
+  "permissions": [ "webNavigation", "<all_urls>", "tabs" ]
```
sublimator, line 35: Could maybe look into build target specific, browser permissions, i.e. only adding the "tabs" permission when Safari is the build target. The "accept-monetization" header implemented via declarativeNetRequest also needs some system for declaring permissions for specific build targets.

PR #1989, opened 2021-06-10
feat(ci): add publish-packages job

### LINKED ISSUE #1960: Setup CI held job to publish `web-monetization-types` to npm
I would be nice if WebMonetization projects outside this repo could make use of types from https://github.com/coilhq/web-monetization-projects/blob/main/packages/web-monetization-types/src/index.ts.

(And maybe leave the extension of `Document` to end-users)

sublimator:
image

Closes #1960 

Add a job to publish web-monetization-types behind an approval job.
Requires NPM_AUTH_TOKEN env var to be set securely in Circle UI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,9 +278,17 @@ workflows:
       - packages-build-scripts
       - hold-publish-packages:
           type: approval
+          filters:
+            branches:
+              only:
+                - main
       - publish-packages:
           requires:
             - hold-publish-packages
+          filters:
+            branches:
+              only:
+                - main
       - jest-all:
           context: web-monetization-tests
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -66,6 +66,29 @@ jobs:
             yarn addons-linter coilfirefoxextension@coil.com.xpi
       - *save_cache
 
+  publish-packages:
+    docker:
+      - image: circleci/node:14-buster-browsers
+    steps:
+      - checkout
+      - *restore_cache
+      - *yarn_install
+      - run:
+          name: Configure NPM authentication
+          command: |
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -66,6 +66,29 @@ jobs:
...
+      - run:
+          name: Build the web-monetization-types package
+          command: |
+            cd packages/web-monetization-types
+            yarn build:ts:verbose
+      - run:
+          name: Publish the web-monetization-types package
+          command: |
+            cd packages/web-monetization-types
+            yarn npm publish --access public
```
sublimator, line 89: image

I find this "stalled" yellow dot annoying somehow as am so used to waiting for the green tick. 

I wonder if using the following `yarn npm publish` option and removing the hold may be better:

```
  --tolerate-republish    Warn and exit when republishing an already existing version of a package
```

traviscrist: Its a really annoying side effect of the hold. You could just auto publish on main by filtering the workflow to only the main branch. 

Another alternative is to set this to run when a new tag is made.

sublimator:
Yeah, main and tolerate flag sounds good?

sublimator:
Though I think u sold me on pushing to a tag formed a certain way

traviscrist:
> Though I think u sold me on pushing to a tag formed a certain way

You could do main and a new tag and then it only runs when that happens. That's how a lot of the ILP NPM packages do it list ILP-STREAM iirc.

sharafian:
I've added a token under the NPM_AUTH_TOKEN field in circleCI with permissions to publish to the webmonetization NPM organization

sublimator:
@traviscrist 
I just put a "main only" filter on the approve job, so will only see the annoying progress yellow dot on main, which we can live with? Day to day PRs will still tick nice and green.

PR #2025, opened 2021-06-17
fix(polyfill): better handle multiple tags

sublimator:


packages/webmonetization-polyfill-utils/src/lib/MonetizationTagObserver.ts
```diff
@@ -82,7 +82,13 @@ export class MonetizationTagObserver {
     const metas: MetaList = this.head.querySelectorAll(
       'meta[name="monetization"],link[rel="monetization"]'
     )
-    metas.forEach(this.onAddedTag.bind(this))
+    metas.forEach(m => {
+      try {
+        this.onAddedTag(m)
```
sublimator, line 87: https://github.com/coilhq/web-monetization-projects/blob/d3c533c84d5565875432fae5db2894b3f4a3305e/packages/coil-extension/src/content/services/ContentScript.ts#L205-L214

Prior to this change (swallowing exceptions and simply logging them, which is essentially what is done when tags are detected dynamically) if there are 2 tags in the DOM already when the above code runs, then the pause/resume handlers will not be set, thus allowing background monetization.

packages/webmonetization-polyfill-utils/src/lib/MonetizationTagObserver.ts
```diff
@@ -158,6 +164,15 @@ export class MonetizationTagObserver {
     }
 
     const details = this.getPaymentDetails(meta)
+
+    if (this.metaTags.size + 1 > this.maxMetas) {
```
sublimator, line 168: I believe at one point in the design phase the code was intended to handle the case of tag 2 starting once tag 1 was removed, which would have been a overwrought, but that may explain why we were tracking the tags in the Map BEFORE throwing errors

sharafian (approved):

PR #2070, opened 2021-06-29
fix(coil-extension): query canTip everywhere and auth tweaks

sublimator:


packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -153,17 +157,12 @@ export class AuthService extends EventEmitter {
       )
       this.trace('after refreshTokenAndUpdateWhoAmi', token)
     } else {
-      // Routinely do a whoami query to check for subscription status
+      // Routinely do a whoami query to check for subscription status or
```
sublimator, line 160: The Coil backend doesn't have GraphQL subscriptions set up for field changes?

sharafian: no, we don't have graphQL subscriptions set up

sharafian (approved):
lgtm, can I try this out in CoilDev to check that it addresses the issue? Could get Dees and Uchi to verify too

PR #2075, opened 2021-06-30
fix(webexts-build-utils): use sourceMaps in dev/transpileOnly mode, closes #2065

### LINKED ISSUE #2065: [Firefox] background page logging numbers are off

sublimator:


packages/coil-anonymous-tokens/test/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
   // FROM UPKEEP TEMPLATE
-  "extends": "../../../tsconfig.json"
+  "extends": "../../../tsconfig.json",
```
sublimator, line 3: want to leave the root tsconfig for the ide, with these types of options disabled (for performance reasons - *assuming* it actually is a perf hit)

PR #2118, opened 2021-07-08
fix(coil-client): fix imports causing undefined whoami selection

sublimator:


packages/coil-client/src/queries/queryToken.ts
```diff
@@ -1,5 +1,7 @@
 import { CoilUser } from '../types'
-import { GraphQlClient, whoamiSelection } from '..'
+import type { GraphQlClient } from '../graphQlClient'
+
+import { whoamiSelection } from './whoAmI'
```
sublimator, line 4: The bug:
At this point whoamiSelection when imported from `..` is `undefined`

sublimator: See:
https://github.com/coilhq/web-monetization-projects/blob/c2d534e1705122014f6c4e74293ba1178b9350c8/packages/coil-client/src/graphQlClient.ts#L33-L37

PR #2301, opened 2021-10-15
feat: tipping closed beta clean

traviscrist:
- All conflicts resolved and rebased on main with test and format failures fixed. 
- Use this for the closed beta tipping feature


packages/coil-extension/src/popup/components/icons/CreditCardIcon.tsx
```diff
@@ -0,0 +1,40 @@
...
+import { Discover } from './Discover'
+import { JCB } from './JCB'
+import { MasterCard } from './MasterCard'
+import { Visa } from './Visa'
+import { CreditCardPlaceholder } from './CreditCardPlaceholder'
+
+//
+// Models
+//
+interface ICreditCardIcon extends SvgIconProps {
```
sublimator, line 15: Why not CreditCardIconProps ?

packages/coil-extension/src/popup/components/views/TipRouter.tsx
```diff
@@ -0,0 +1,76 @@
+/*
+    TipRouter
+    
+    The tip router is responsible for all the local state that is relative the the tipping process as well as the logic that determines which tipping view is rendered. Also will need to pass context down to the rendered components.
```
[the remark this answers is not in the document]
mwill8886: sounds good, i'll do that on my branch.

packages/coil-extension/src/popup/components/icons/CreditCardIcon.tsx
sublimator: I was actually surprised they got through, as actually @typescript-eslint used to ban that form (I used to use it too, though I adopted it from Java). It must be allowed these days by the default rules

PR #2424, opened 2021-12-04
feat: initial impl of WM2

sublimator:
Backwards compatible implementation of https://github.com/WICG/webmonetization/pull/193

- [ ] Get payment pointer for tipping from the TagManager, first "active" one in document 
   - currently it could be the case that pages with mixed meta/link will not work 
      - [ ] Why does coil-wordpress-plugin use two tags? there hasn't been an extension that *doesn't* support link tags for 
              long time?
- [ ] SPSP link load events when user isn't logged in, currently these won't be emitted as there's no dedicated SPSP 
       service. 
- [x] Tipping works when extension loaded as a temporary add-on (production or dev build) but not when loaded from add-on store 
  - this also seems to be the case for Chrome CoilPreview as well - 
     - [x] at least with a page with a body link
    - [x] try simple meta in head page
- [ ] Firefox CSP issue, ~~seemingly only when add-on is published~~ [publishing](https://addons.mozilla.org/en-GB/firefox/addon/coilwm2preview/): image
  - [x] verify it's not happening when loaded as a local add-on 
     - it IS happening locally 
- [x] Monetization seems to be globally declared somehow, supposed to be block scoped, maybe need wrap everything in a function
  - [x]  check if it's also the case on Chrome, so far only seen it on FF
    - actually, it's also the case with Chrome
- [x] Create a cloneInto helper for the CustomEvent#detail on FF, remove copy/pasted cloneIntoRef code
- [x] Route monetization events to link
  - [x] **PRIORITY:HIGH** `amount` should be a bigint as per https://github.com/WICG/webmonetization/pull/193/commits/3d5e8f1dd01ee7f3e606c178977400df7005e341#diff-772384ebc24069774f1ac50dd3b8ba92c488eaca9a0029ce4aa0c3f65ce86de8R632
- [x] Enforce back-compat meta in head
- [x] **PRIORITY:HIGH** make sure tipping "works", just use any stream
  - [ ] use the first link in the document ? create multiple tips ?
- [x] Support `@onmonetization` attr
  - [x] Window
  - [x] HTMLElement
  - [x] Document  (@marcoscaceres agreed it made sense)
- [x] Support `@disabled` checking on link
- [x] Test iframes
  - [x] test WM1 iframes 
  - [ ] test WM2 iframes
- [x] Log deprecation warnings for <meta> tags ? By default ? Behind a flag?
  - only when localStorage.WM_DEBUG is enabled
- [x] Make sure link tags only accept urls, not `$payment.pointers/k-thx`
- [x] Support links in body
- [x] **PRIORITY:HIGH** **BUG** as reported by @SarahCoilAccount, see: https://coiltech.slack.com/archives/D028151DA9M/p1645430884957179, icon doesn't return to fully lit green after re-activating a WM tab
  - [x] Fix: image
- [x] **PRIORITY:HIGH** test on FF
  - [x] **BUG**  The Event subclass doesn't seem to be work on FF  `class MonetizationEvent extends Event {`
     - [x] **PRIORITY:HIGH** investigate just wrapping addEventListener, capturing a CustomEvent in a listener wrapper, then sending a MonetizationEvent to the wrapped function
     - [x] Fixed with cloneInto to share custom event detail object @ f15454e9af1048e7361c2eb518448ca7f1b3553e, thanks to @[marcoscaceres](https://github.com/marcoscaceres)
- [ ]  Listen to the play and pause events, and if any audio source emits a play event there’s audio playing until it emits a pause event ([ppk](https://github.com/pp-koch))
  - see: https://quirksmode.org/coil/testsuite/audio.html
  - see: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes
- [ ] **PRIORITY: HIGH** localStorage.WM_DEBUG ([ppk](https://github.com/pp-koch))
   - [ ] No monetization link - ( nd: log no link on document ready ? what about tags added dynamically a touch slowly? ) 
   - [x] No valid URL in link
   - [x] JSON file not found
   - [x] JSON file malformed
   - [x] JSON loaded correctly
   - [x] SPSP error (not sure if that exists or you can read it out)
   - [ ] No reply from Interledger after SPSP request
   - [ ] And maybe more on the Interledger side
- [x] restart streaming after network drops ([ppk](https://github.com/pp-koch)) 
  - tested and it restarts like it used to do
- [ ] Implement sourceAsset, sourceAmount, sourceScale
  - See: https://github.com/WICG/webmonetization/pull/193#issuecomment-1008925303
  - [ ] Marcos is not keen on this for privacy reasons (TODO: clarify)
    - @adrianhopebailie thinks it's not an issue
- [x] Support multiple links
  - [ ] **PRIORITY:HIGH** Verify whether multiple streams share the same active token
  - [x] **PRIORITY:HIGH** Support `meta tag, then link tag` in document (ala WM wordpress)
     - currently gets in a weird state
     - image
     - [x] move `this.assoc.addStreamId(frame, requestId)` higher up in the startWebMonetization method
  - [x] on iframe allow changed needs to be at least sensible, either not supported (marcos said browsers ignore 
         the change after framed document has loaded ) or working with all the tags in a page
  - [x] pauseWebMonetization on idle/visibilityState change will pause multiple streams
  - [x] resumeWebMonetization on idle/visibilityState change will resume multiple streams
  - [x] tabStates[tabId][frameId].lastMonetization assumes 1:1 frame:requestId
     - [x] Need a better way to detect stale requestIds 
     - [ ] Track state in monetization service
  - [x] StreamAssociations will need to store a Set<string> rather than string for a given [tabId][frameId]
  - [x] browserAction icon badges now need logic from more than one "total" 
  - [ ] what else?? 
  - [ ] All the TODO:WM2 entries need auditing/fixing
  - [x] FrameState['monetized'] should be set properly by stopWebMonetization 
  - [x] one stream abort will take down all streams
- [x] Window.prototype.onmonetization setter 
- [x] HTMLElement.prototype.onmonetization setter
- [ ] CSP `monetization-src`
   - It will require permissions to access the response headers 
- [x] link.rellist.supports('monetization') (https://github.com/WICG/webmonetization/issues/246#issuecomment-994596405)
- [x] force a runtime.reload upon notification button acceptance (*might* do an update??)
  - [ ] didn't work, try the api chrome.runtime.requestUpdateCheck()
- [ ]  make sure SPSP is "not delayed"
- [x] Route SPSP (load/error) events to link tag
  * [x] what about loadstart/abort
    - marcos says need a valid/defined use case to justify events
  * [x] what about multiple load/error events in cases of retry ? 
    * Might need to think of the events as an *initial* validation phase, with subsequent retries being "protocol layer" stuff
   - [x]  "when I change the href of a link to a page that does not exist the load event still fires, while I’d expect the error event" ([ppk](https://github.com/pp-koch))
      - [x] is this because the stream is still firing internal events due to having not winded down properly yet ? 
- [ ] Implement separate background SPSP service
- [x] document.monetization.state,monetizationxxxx events in case of one link tag (backwards compatibility)
  - [x] only set state, emit events when one tag (link or meta) and it's in the head
  - [x] Pass the old tests @ https://wm.niqhome.com/


packages/coil-extension/src/background/services/StreamAssociationsWM2.test.ts
```diff
@@ -0,0 +1,90 @@
+import '@abraham/reflection'
```
sublimator, line 1: wonder if we can move this to some jest config, rather than have to repeat it everywhere ??

packages/coil-extension/src/background/services/StreamAssociationsWM2.ts
```diff
@@ -0,0 +1,68 @@
...
+    return this.streamsToFrames[streamId]
+  }
+
+  setStreamFrame(streamId: string, frame: FrameSpec): void {
+    this.streamsToFrames[streamId] = frame
+  }
+
+  clearStreams({ tabId, frameId }: FrameSpec): void {
+    this.getStreams({ tabId, frameId }).forEach(
+      this.clearStreamFrame.bind(this)
```
sublimator, line 48: todo fix

packages/coil-extension/src/background/services/MonetizationService.ts
```diff
@@ -417,4 +411,21 @@ export class MonetizationService {
     }
     this.api.tabs.sendMessage(tabId, message, { frameId })
   }
+
+  handleStreamsAbortEvent() {
+    this.streams.on('abort', requestId => {
+      this.log('aborting monetization request', requestId)
+      const frame = this.assoc.getStreamFrame(requestId)
+      if (frame) {
+        // TODO: one stream abort will take down all streams??
```
sublimator, line 420: needs TODO:WM2

packages/coil-extension/src/background/background.ts
```diff
@@ -71,8 +36,31 @@ async function main() {
...
-  window.bg = container.get(BackgroundScript)
+  await configureContainer({
+    container: container,
+    loggingEnabled,
+    coilDomain: COIL_DOMAIN,
+    wextApi: API,
+    buildConfig: BUILD_CONFIG,
+    storage: localStorage,
+    getActiveTab: async () => {
+      // This query will not pick up dev tools tabs which may be currently active
```
sublimator, line 47: We could just delay startup until the active Tab is found

PR #2432, opened 2021-12-07
feat: new UI navbar

mwill8886:
resolves: COIL-1565
resolves: COIL-1569
resolves: COIL-1571

- Adds a temporary `NewExtention` component to build out the new UI
- Created `NavBar` component 
- Added a routing context to maintain the state of what page the UI is on
- Created new `Header` component 
- Created new `SettingsView`
~~- Sets default tip amount to users last tipped amount or 1~~
- Wrapped the MUI `createMuiTheme` with wrapper in order to add new props to theme


packages/coil-extension/src/popup/Index.tsx
```diff
@@ -7,6 +7,7 @@ import { PopupHost } from './types'
 import { PopupHostContext } from './context/popupHostContext'
 import { StoreContext, useStoreState } from './context/storeContext'
 import { ViewRouter } from './components/views/ViewRouter'
+import { RouterProvider } from './context/routerContext'
```
sublimator, line 10: Should we update the extension to material ui 5 while doing this? 
Might be a good chance ?

mwill8886: That's a good thought, I don't think there are any breaking changes but I'll check the docs before updating. If it's all good then an update will be a separate PR.

davebarr (approved):

packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -228,11 +228,14 @@ export class AuthService extends EventEmitter {
         inTippingBeta: featureFlagResp.data.featureEnabled,
         minimumTipLimit: minTipLimitResp.data.minTipLimit.minTipLimit
           ? Number(minTipLimitResp.data.minTipLimit.minTipLimit) / 100
-          : 1,
+          : 1, // convert from cents to dollars
         remainingDailyAmount: this.store.user?.tipping?.limitRemaining
           ? Number(this.store.user?.tipping?.limitRemaining) / 100
           : 0, // convert from cents to dollars
-        hotkeyTipAmounts: [5, 10, 50] // dollar amounts - not yet set by user
+        lastTippedAmountUSD: this.store.user?.tipping?.limitRemaining
```
traviscrist, line 235: nit: This math should really be a share util function since we are doing it all over the place.

packages/coil-extension/src/types/user.ts
```diff
@@ -1,7 +1,10 @@
 export interface User {
+  newUi?: boolean // todo: remove this when testing the new ui is done
```
traviscrist, line 2: Is the effectively a feature flag? Maybe we should call this like `newUi2022` or `newUiV2` so we have more context. `newUiTipping`

traviscrist (approved):

packages/coil-extension/src/types/user.ts
mwill8886: There's no real data behind it or db entry for it. It's just a variable I flip when working on it, i'll be the only one using it. Once the UI is updated everyone will get the new UI and the tipping will still be behind the `tipping-beta` feature flag.

packages/coil-extension/src/background/services/AuthService.ts
mwill8886: I actually really want to remove this section and do it all better, I just don't know extactly what I wanted to do just yet. I figured I would match what I had there for now and make the refactor on a new branch.

chris-jackson-actionqa (changes requested):
Running through some test scenarios with @mwill8886 , and we hit an issue.

The popup doesn't list the last tip amount unless you refresh your session.  Mike saw this and realized the data isn't getting rehydrated.  We discovered this when hitting the max tip limit.  If you don't refresh the page, the popup doesn't say anything about your tip limit.  But when I switched to my account and refreshed my coil settings account page, and then went back to the heroku app, then the limit was showing as reached.

So, you need to manually refresh your session in order to get your tip data to show up properly in the extension (and on the coil settings account tipping tab).  

@mwill8886 is investigating.

sublimator:
That rings a bell.
I recall in a previous iteration of tipping, having to trigger another
whoami query to update the tipping balance etc

packages/coil-extension/src/types/user.ts
traviscrist: kk, that makes sense to just leave it then. 👍

davebarr (approved):

mwill8886:
> Running through some test scenarios with @mwill8886 , and we hit an issue.
> 
> The popup doesn't list the last tip amount unless you refresh your session. Mike saw this and realized the data isn't getting rehydrated. We discovered this when hitting the max tip limit. If you don't refresh the page, the popup doesn't say anything about your tip limit. But when I switched to my account and refreshed my coil settings account page, and then went back to the heroku app, then the limit was showing as reached.
> 
> So, you need to manually refresh your session in order to get your tip data to show up properly in the extension (and on the coil settings account tipping tab).
> 
> @mwill8886 is investigating.

Upon investigation it will update if any tab in the browser session reloads. However this does effect more areas than the last tipped amount and some of the features on the new ui (that haven't been built yet) will be effected. I'm making a new linear issue to refactor how we grab the tipping info so that it gets called everytime we use the tipping views.

chris-jackson-actionqa (approved):

PR #2445, opened 2021-12-14
feat(coil-extension): use workspace angular/core with spec ^0.0.0

sublimator:


packages/coil-extension/src/content/content.ts
```diff
@@ -38,3 +39,5 @@ function main() {
   container.get(ContentScript).init()
 }
 main()
+
+console.log({ YARN_PRIORITY })
```
sublimator, line 43: Yeah, yarn will default to the workspace version

PR #2480, opened 2022-01-06
feat: update puppeteer to 16.0.1

sublimator:


sublimator:
There still seems to be a problem on node 16 with about:blank

chris-jackson-actionqa (approved):

PR #2484, opened 2022-01-07
feat: new ui streaming views

mwill8886:
resolves: COIL-1570

The new UI is still not visible to the public and will be behind a feature flag
Added the streaming views to the new extension UI.
Streaming View States: 
- Web monetized site
- Not web monetized site
- User with no membership
- Coil site
- Coil site, discover page


packages/coil-extension/src/popup/contants.ts
```diff
@@ -1,9 +1,10 @@
...
   streaming = 'streaming',
-  streamingNoWebMo = 'streaming/notmonetized',
-  streamingNoMembership = 'streaming/nomembership',
-  streamingCoil = 'streaming/coil',
-  streamingCoilDiscover = 'streaming/coil/discover',
+  // streamingWebMo = 'streaming/monetized',
+  // streamingNoWebMo = 'streaming/notmonetized',
+  // streamingNoMembership = 'streaming/nomembership',
+  // streamingCoil = 'streaming/coil',
+  // streamingCoilDiscover = 'streaming/coil/discover',
```
bboynton97, line 7: Did you mean to keep these commented out?

packages/coil-extension/src/popup/mocks/loadMockedStates.tsx
```diff
@@ -146,7 +146,8 @@ const userCanTip = mockState({
 
 const userCanTipNewUi = mockState({
   monetized: true,
-  coilSite: 'https://coil.com/p/mirrae/Letter-to-my-Daughter/-ZInTJqN-',
```
bboynton97, line 149: Same here

bboynton97 (changes requested):

packages/coil-extension/src/popup/mocks/loadMockedStates.tsx
mwill8886: This was intentional, `loadMockedState` is the mocked data for the view previewer. That's essentially what I use to build the views.

packages/coil-extension/src/popup/contants.ts
mwill8886: Nope, and I just realized the file is misspelled

packages/coil-extension/src/popup/components/views/Router.tsx
```diff
@@ -38,70 +46,46 @@ export const Router = () => {
...
-            no webmo
-          </button>
-        </NewHeaderFooterLayout>
-      )
+      // check the logic for what state to display for the streaming page
+
+      // Streaming - No Membership
+      if (
+        !user?.subscription ||
+        (user.subscription && !user.subscription.active)
```
sublimator, line 66: Could be condensed:
!user?.subscription?.active

packages/coil-extension/src/popup/components/NavBar.tsx
```diff
@@ -3,7 +3,7 @@ import { Box, styled, Theme } from '@material-ui/core'
 import SettingsIcon from '@material-ui/icons/Settings'
 
 import { useRouter } from '../context/routerContext'
-import { ROUTES } from '../contants'
+import { ROUTES } from '../constants'
```
bboynton97, line 6: love it

bboynton97 (approved):

chris-jackson-actionqa (approved):
Approving, but with reservations.  Not able to test, so not sure if things are correct.  But we need this checked in order to actually test it.  Catch 22.

Issue #2087, opened 2021-07-02
Pre push hook is busted on MacOS Monterey


sublimator:
```
➜  coil-extension git:(nd-update-safari-with-ios-support-2021-07-02) ✗ gpsup
Pre-push Hook: Checking branch name
parent_pid: 27901
parent_pid: ps -ocommand= -p 27901
ps: option requires an argument -- p
usage: ps [-AaCcEefhjlMmrSTvwXx] [-O fmt | -o fmt] [-G gid[,gid...]]
          [-g grp[,grp...]] [-u [uid,uid...]]
          [-p pid[,pid...]] [-t tty[,tty...]] [-U user[,user...]]
       ps [-L]
^C
```

sublimator:
Requires git to be on $PATH
("Fixed" in my case by a `brew install`)

Need to fix loop to find git in process tree:
```
until [[ "$push_command" == git* ]]; do
```
Must handle `/Some/long/path/to/git`

bboynton97:
Is there a particular reason why we are protecting branches with a pre-push script rather than just using github's branch protection feature?

bboynton97:
For others running into this issue, I had two git installations:
- `/usr/bin/git`
- `/opt/homebrew/bin/git`

The first installation was failing on this pre-push script. Upon switching to use the second installation, the pre-push script passed.

sublimator:
see also:
https://github.com/coilhq/web-monetization-projects/pull/2490/commits/40d1330f55e9d324bc84885d20f377e6e6fc9176#r782643437

traviscrist:
@sublimator I can't recall why we settled on it being this way. I think it may have been since it worked with your local dev flow better @sublimator Could we run this in CI and then just setup branch protection?

PR #2490, opened 2022-01-10
feat(extension): fetch last tip

bboynton97:
Closes: COIL-1642

Refactors tipping logic into `TippingService` and a formatters file.

Retrieves the latest tip settings after a successful tip.

### DIFF
--- a/packages/coil-extension/src/background/services/AuthService.ts
+++ b/packages/coil-extension/src/background/services/AuthService.ts
@@ -6,6 +6,7 @@ import { inject, injectable } from 'inversify'
 import { LocalStorageProxy } from '../../types/storage'
 import * as tokens from '../../types/tokens'
 import { TimeoutError } from '../../util/timeout'
+import { formatTipSettings } from '../util/formatters'
 
 import { SiteToken } from './SiteToken'
 import { Logger, logger } from './utils'
@@ -182,16 +183,24 @@ export class AuthService extends EventEmitter {
 
     this.log('updateWhoAmi resp', resp.data)
     if (resp.data?.whoami) {
-      this.store.user = {
-        ...resp.data.whoami
-      }
+      this.store.user = resp.data.whoami
 
       // Data needed for tipping
       // tipping-beta: featureEnabled: boolean
       // minimum tip limit: minTipLimit > minTipLimit
       // remaining daily amount: whoami > tipping > limitRemaining
 
-      this.formatTipSettings(token)
+      if (this.store.user) {
+        this.store.user = {
+          ...this.store.user,
+          tipSettings: await formatTipSettings(
+            this.client,
+            token,
+            this.store.user?.tipping?.limitRemaining,
+            this.store.user?.tipping?.lastTippedAmount
+          )
+        }
+      }
       return token
     } else {
       return null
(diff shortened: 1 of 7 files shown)


commands/pre-push
```diff
@@ -11,11 +11,11 @@ REFS=$(cat)
 parent_pid=$PPID
 push_command=''
 
-until [[ "$push_command" == git* ]]; do
-  push_command=$(ps -ocommand= -p $parent_pid)
-  parent_pid=$(ps -o ppid= $parent_pid)
-  # echo "command = $push_command"
-done
+#until [[ "$push_command" == git* ]]; do
```
sublimator, line 14: https://github.com/coilhq/web-monetization-projects/issues/2087

packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -182,16 +183,24 @@ export class AuthService extends EventEmitter {
...
       // Data needed for tipping
       // tipping-beta: featureEnabled: boolean
       // minimum tip limit: minTipLimit > minTipLimit
       // remaining daily amount: whoami > tipping > limitRemaining
 
-      this.formatTipSettings(token)
+      if (this.store.user) {
+        this.store.user = {
+          ...this.store.user,
+          tipSettings: await formatTipSettings(
```
mwill8886, line 196: I'm wondering if instead of having the `formatTipSettings` here can we make the call to the tipping service for `updateTipSettings`?

packages/coil-extension/src/background/util/formatters.ts
```diff
@@ -0,0 +1,23 @@
...
+  client: GraphQlClient,
+  token: string,
+  limitRemaining: any,
+  lastTippedAmount: any
+) {
+  // convert all tip settings from cents to dollars
+  // set default hotkey tip amounts since we don't yet get them from the user
+  // add feature flag and minTipLimit
+  const featureFlagResp = await client.featureEnabled(token, 'tipping-beta')
+  const minTipLimitResp = await client.minTipLimit(token)
```
mwill8886, line 13: I know this is how I did it originally and I'm sorry :( Can we remove the api calls from the formatter and just add them to the `tipSettings` query, then only have the formatter util format the data?

packages/coil-extension/src/background/services/AuthService.ts
traviscrist: @mwill8886 whats the pro vs con of doing that? Is it just one more api call but the data will be cleaner? Do you need a backend update for this?

commands/pre-push
```diff
@@ -11,11 +11,11 @@ REFS=$(cat)
 parent_pid=$PPID
 push_command=''
 
-#until [[ "$push_command" == git* ]]; do
-#  push_command=$(ps -ocommand= -p $parent_pid)
-#  parent_pid=$(ps -o ppid= $parent_pid)
-#  # echo "command = $push_command"
-#done
+until [[ "$push_command" == git* ]]; do
```
sublimator, line 14: I had a look at this again and this section you commented out is all just debugging variables really.
It was a script I adapted from some [online source](https://stackoverflow.com/a/63322932/11587288)

> The script basically will do 2 things:
This script will block anybody who tries to push to a certain branch (in my case I don't want anybody -including myself- to push directly to master and develop branch). They need to work in their own branch and then create a pull request.
This script will block anybody who tries to push to a branch that is different from their current active branch. For example you are in branch fix/someissue but then you mistakenly type git push origin master.


"Why use this over protected branches"
Good question. I can't think of anything at present

Although I have dim memories of discussing this with @traviscrist. 
It may have been something to do with the way our PR merge permissions are configured and some Github inflexibility (which may or may not exist any more)

bboynton97: I got the script working on my local machine and added how to the issue you linked previously. I still think we might consider protected branches, but I'm okay with leaving this how it is currently.

mwill8886 (approved):

PR #2535, opened 2022-01-21
feat: update images copy and fixes

mwill8886:
resolves: COIL-1678
resolves: COIL-1693
resolves: COIL-1700
resolves: COIL-1711

- added support for lottie
- updated graphics and animations
- streaming on webmo'd page animation will only play if it is streaming
- updated streaming copy to reflect content and sites (handles youtube and twitch)
- added font-family fallback


packages/coil-extension/tsconfig.build.json
```diff
@@ -8,7 +8,8 @@
     "rootDir": "src"
   },
   "include": [
-    "src"
+    "src",
+    "src/popup/components/lottie-animations/*.json"
```
sublimator, line 12: This is file is automatically generated by upkeep

chris-jackson-actionqa (approved):

PR #2552, opened 2022-01-25
feat: fix max tip amount

mwill8886:
resolves: COIL-1712

This fixes the issue of the tipping limit not being set to $0 on when a user switches from a webmo'd tab to a non webmo'd tab.

### DIFF
--- a/packages/coil-extension/src/popup/context/tipContext.tsx
+++ b/packages/coil-extension/src/popup/context/tipContext.tsx
@@ -2,6 +2,7 @@ import React, { createContext, useContext, useState, useMemo } from 'react'
 import { StorageService } from '@webmonetization/wext/services'
 
 import { getCreditCardFromPaymentMethods } from '../../util/getCreditCardFromPaymentMethods'
+import { calculateMaxAllowableTip } from '../../util/calculateMaxAllowableTip'
 
 //
 // Models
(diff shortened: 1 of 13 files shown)


packages/coil-extension/src/popup/context/tipContext.tsx
```diff
@@ -1,40 +1,100 @@
...
   currentTipAmount: number
+  maxAllowableTipAmount: number //* the maxAllowableTip is primarily responsible for disabling tipping inputs
   setCurrentTipAmount: (amount: number) => void
 }
+
 interface ITipProvider {
   storage: Pick<StorageService, 'get'>
 }
 
+//
```
chris-jackson-actionqa, line 20: This is duplicate code of what's in the utils directory.

packages/coil-extension/src/util/calculateMaxAllowableTip.ts
```diff
@@ -0,0 +1,29 @@
+export const calculateMaxAllowableTip = (
```
chris-jackson-actionqa, line 1: If not already done, this needs unit tests.

chris-jackson-actionqa (changes requested):

njlie (approved):
LGTM aside from Chris' comments, approving so as not to block on those.

packages/coil-extension/src/popup/context/tipContext.tsx
mwill8886: Yeah i was in the process of adding the tests and moving the calc to a util when i had to fix something for the demo

packages/coil-extension/src/util/calculateMaxAllowableTip.ts
mwill8886: already done 👍

chris-jackson-actionqa (approved):

PR #2555, opened 2022-01-27
feat(extension): update tip credit requests

njlie:
Part of the work for: COIL-1698.
* Updates GraphQL calls to use `totalTipCredit` on the `User` object as the Tip Credit balance.

note: tests will fail until the new API schema for tip credits is deployed to staging. See https://github.com/coilhq/coil/pull/5270


njlie:
Superseded by https://github.com/coilhq/web-monetization-projects/pull/2568, closing.

PR #2572, opened 2022-02-03
feat: changed behavior of hotkeys to no longer forward step

mwill8886:
resolves: COIL-1730

Updates the behavior of the hotkeys so that they no longer will automatically forward the step once clicked. They will update the tip amount only.

Also split email and domain on settings page if no shortname and fullname exist


packages/coil-extension/src/popup/components/views/SettingsView.tsx
```diff
@@ -95,18 +95,21 @@ export const SettingsView = () => {
...
   const getSecondaryName = () => {
     if (user?.fullName && user?.shortName) {
       return user.shortName
     } else {
-      if (user?.shortName) {
+      if (user?.shortName || user?.fullName) {
         return user.email
+      } else {
+        const emailDomain = user.email.split('@')[1]
+        return `@${emailDomain}`
```
chris-jackson-actionqa, line 111: So, secondary name can be "gmail.com"?  lol!

chris-jackson-actionqa (approved):

packages/coil-extension/src/popup/components/views/SettingsView.tsx
mwill8886: Yup, in this instance is will be MyEmailAddress for the top line and @Domain.com on the second line. I'm splitting the email in the event that there is no shortName and no fullName. This will help a little with really long emails while keeping the same aesthetic.

davebarr (approved):

PR #2603, opened 2022-02-14
feat: added feedback for daily limit being below min tip

mwill8886:
resolves: COIL-1842

If the remaining daily limit is less than the minimum tip limit, show updated feedback for user.

### DIFF
--- a/packages/coil-extension/src/popup/components/AmountInput.tsx
+++ b/packages/coil-extension/src/popup/components/AmountInput.tsx
@@ -195,7 +195,10 @@ export const AmountInput = (): React.ReactElement => {
         <Amount
           size={displayFontSize}
           onClick={() => setIsUserInput(true)}
-          disabled={maxAllowableTipAmountUsd == 0}
+          disabled={
+            maxAllowableTipAmountUsd == 0 ||
+            maxAllowableTipAmountUsd < minTipLimitAmountUsd
+          }
         >
           $
           {Number.isInteger(currentTipAmountUsd)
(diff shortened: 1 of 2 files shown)


bboynton97 (approved):

packages/coil-extension/src/popup/components/AmountInput.tsx
```diff
@@ -195,7 +195,10 @@ export const AmountInput = (): React.ReactElement => {
         <Amount
           size={displayFontSize}
           onClick={() => setIsUserInput(true)}
-          disabled={maxAllowableTipAmountUsd == 0}
+          disabled={
+            maxAllowableTipAmountUsd == 0 ||
+            maxAllowableTipAmountUsd < minTipLimitAmountUsd
```
davebarr, line 200: do these map to the same case?  can minTipLimitAmountUsd be 0? (does that even make sense?)

davebarr (approved):

packages/coil-extension/src/popup/components/AmountInput.tsx
sublimator: I think you're right in that  "maxAllowableTipAmountUsd < minTipLimitAmountUsd" will handle all "maxAllowableTipAmountUsd == 0" but the redundancy does make it a bit more readable? 

I wondered if this calculation should be added to the `calculateMaxAllowableTip` but then I guess it makes what's happening in TipAmountFeedback a bit trickier.

PR #2616, opened 2022-02-17
feat(extension): allow users without memberships to tip

njlie:
Allows users without a membership to tip if they have tip credits. This is accomplished by maintaining a per-iframe (which is further grouped by tab) object that maps payment pointers to them. This is similar to the way that the extension maps ILP streams to iframes grouped by tab. 

This allows the `tip` command to complete without relying on the existence of an ILP stream to get the payment pointer it needs to send to.

Closes COIL-1831.


packages/coil-extension/src/content/services/ContentScript.ts
```diff
@@ -76,14 +76,31 @@ export class ContentScript {
       this.runtime.sendMessage(request)
     }
 
+    const associatePaymentPointer = (paymentPointer: string) => {
```
sublimator, line 79: We ought to be able to piggy back the existing start/stop commands though?
there shouldn't be need for any new messages

sublimator:
will revisit this for next release, see also https://github.com/coilhq/web-monetization-projects/pull/2617
https://coiltech.slack.com/archives/C01Q12WE5MG/p1645160293161159?thread_ts=1645156942.198099&cid=C01Q12WE5MG

bboynton97 (approved):

njlie:
Pulled in changes from https://github.com/coilhq/web-monetization-projects/pull/2617. Thanks @sublimator!

PR #2635, opened 2022-02-23
feat: update icon for tipping only

mwill8886:
resolves: COIL-1566

Updates the icon for the browser to show the green circle outline on a monetized page for non-members who should be allowed to see the new UI and tip.


packages/coil-extension/src/background/services/BackgroundScript.ts
```diff
@@ -362,6 +364,18 @@ export class BackgroundScript {
                 ? 'streaming'
                 : 'streaming-paused'
             this.tabStates.setIcon(tabId, state)
+          } else {
+            // need to check if the user user is able to tip with the new ui
```
bboynton97, line 368: nit of a nit: `user user`

bboynton97 (approved):

PR #2636, opened 2022-02-24
feat: disallow fractional inputs

mwill8886:
resolves: COIL-1920

limits manual input of tips to whole numbers and does not allow users to input fractional amounts.

### DIFF
--- a/packages/coil-extension/src/popup/components/TipAmountFeedback.tsx
+++ b/packages/coil-extension/src/popup/components/TipAmountFeedback.tsx
@@ -77,6 +77,7 @@ export const TipAmountFeedback = () => {
     return null
   }
 
+  // This formatting is here in case we choose to start allowing fractional amounts later on
   const formattedTipCreditAmountUsd = Number.isInteger(totalTipCreditAmountUsd)
     ? totalTipCreditAmountUsd
     : totalTipCreditAmountUsd.toFixed(2)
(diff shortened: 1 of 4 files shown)


packages/coil-extension/src/popup/components/TipAmountFeedback.tsx
```diff
@@ -76,13 +77,17 @@ export const TipAmountFeedback = () => {
     return null
   }
 
+  const formattedTipCreditAmountUsd = Number.isInteger(totalTipCreditAmountUsd)
+    ? totalTipCreditAmountUsd
+    : totalTipCreditAmountUsd.toFixed(2)
```
traviscrist, line 82: Do we need to keep the toFixed(2) here?

traviscrist: Ooh I see this lets us do the decimal later if we want to... should we add a comment about why this is kept here?

mwill8886: I think the comment is a good idea, I just added a comment for this.

traviscrist: Looks good, thx!

traviscrist (approved):

PR #2652, opened 2022-03-03
feat: keyboard accessible buttons

mwill8886:
resolves: COIL-1889

Added keyboard accessiblility for the buttons and input on the extension

### DIFF
--- a/packages/coil-extension/src/popup/components/TipAmountFeedback/RestrictedMessage.tsx
+++ b/packages/coil-extension/src/popup/components/TipAmountFeedback/RestrictedMessage.tsx
@@ -39,7 +39,11 @@ export const RestrictedMessage = () => {
       return (
         <Typography variant='subtitle1'>
           Max tip amount.{' '}
-          <LinkUnderlined onClick={tabOpener(`${coilDomain}/settings/tipping`)}>
+          <LinkUnderlined
+            tabIndex={0}
+            onKeyPress={tabOpener(`${coilDomain}/settings/tipping`)}
+            onClick={tabOpener(`${coilDomain}/settings/tipping`)}
+          >
             Raise limit
           </LinkUnderlined>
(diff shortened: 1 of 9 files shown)


chris-jackson-actionqa (commented):
I wonder if we can add a test to the testing/e2e-extension suite?  Just some simple keyboard interactions.

packages/coil-extension/src/popup/components/TipAmountFeedback/RestrictedMessage.tsx
```diff
@@ -39,7 +39,12 @@ export const RestrictedMessage = () => {
       return (
         <Typography variant='subtitle1'>
           Max tip amount.{' '}
-          <LinkUnderlined onClick={tabOpener(`${coilDomain}/settings/tipping`)}>
+          <LinkUnderlined
+            tabIndex={0}
```
sublimator, line 43: Could this LinkUnderlined thing be a component? 

eg.  `<CoilLink path='/settings/billing'>`

sublimator:
I built this and tested it, and I don't seem to be able to tab to the main amount and edit it (as I can when I click on it)

sublimator:
Given `escape` already closes the popup, having the first tab-able element on the page be the `X` to close the popup seems a waste?

sublimator (approved):

mwill8886:
> Given `escape` already closes the popup, having the first tab-able element on the page be the `X` to close the popup seems a waste? 

Good point, I'll see about editing that tomorrow.

packages/coil-extension/src/popup/components/TipAmountFeedback/RestrictedMessage.tsx
mwill8886: Yeah I think so, I'll update.

Issue #2048, opened 2021-06-23
[Firefox] Firefox logging out from extension


sublimator:
> got signed out of my extension today. on the latest version (loaded as a temporary extension in FF). 

Could FF still be having problems with the /handler.html CSP ?
Unable to get token from coil.com on startup so the extension logs itself out ? 

Note, this could be happening on Chrome too, and the fact it was FF is irrelevant 

Investigate:
- [ ] Containers extension
- [ ] about:profiles usage
- [ ] "built in" containers ?
- [ ] Clearing cookies / site data upon exit

sublimator:
Checked with MacOS Mozilla Firefox (89.0.1) and the background page is successfully getting the token via handler.html.

sublimator:
![image](https://user-images.githubusercontent.com/525211/123401868-64259d80-d5d1-11eb-8a27-2b1e824cced1.png)

natanlao:
I can reproduce this; Coil (just recently) requires sign-in whenever I quit Firefox. I'd prefer to not do this, since going through the auth flow every day / every other day can get a little tiresome.

* Coil 0.0.54
* Firefox 89.0.1 (64-bit)

sublimator:
@natanlao 
Thanks very much for the data point!

sublimator:
@natanlao 
If I'm able to publish a preview version of the extension, with some extra logging circa authentication, would you mind running it and sharing results ? I've personally not been able to reproduce this, despite a few reports from others.

Can you share what platform you're on ? 
How often do you quit Firefox ? You need to sign back in every time? Are you able to reproduce this consistently just by quitting ?

natanlao:
>  If I'm able to publish a preview version of the extension, with some extra logging circa authentication, would you mind running it and sharing results ? I've personally not been able to reproduce this, despite a few reports from others.

Sure!

> Can you share what platform you're on ?

I'm running NixOS:

```
 - system: `"x86_64-linux"`
 - host os: `Linux 5.10.45, NixOS, 21.05.1268.21b696caf39 (Okapi)`
 - multi-user?: `yes`
 - sandbox: `yes`
 - version: `nix-env (Nix) 2.3.12`
```
>  How often do you quit Firefox ?

I don't quit Firefox manually, so once a day when I shut off my PC.

> You need to sign back in every time? Are you able to reproduce this consistently just by quitting ?

[...]

PR #2664, opened 2022-03-07
fix(coil-extension): experimental fix for FF logout problem #2048

sublimator:


packages/coil-extension/manifest.json
```diff
@@ -32,7 +32,7 @@
       "run_at": "document_start"
     }
   ],
-  "permissions": [ "webNavigation", "<all_urls>" ],
+  "permissions": [ "webNavigation", "<all_urls>", "tabs" ],
```
sublimator, line 35: Though mostly likely the latter, I am not sure if it's simply the addition of the `tabs` permission or the different token retrieval method that is the "fix" here

sublimator: It actually may be the case that it *IS* simply the tabs permission:
https://github.com/coilhq/web-monetization-projects/issues/2048#issuecomment-1065740650

Which call?

PR #2679, opened 2022-03-14
fix(coil-extension): use tabs permission on firefox, closes #2048

### LINKED ISSUE #2048: [Firefox] Firefox logging out from extension

sublimator:


legowerewolf:
When might this land in the regular extension in on Firefox?

PR #2717, opened 2022-04-01
feat: set default view to tip view if allowed

mwill8886:
resolves: COIL-1983

set the default extension view to tipping if the user is allowed to tip

### DIFF
--- a/packages/coil-extension/src/popup/context/routerContext.tsx
+++ b/packages/coil-extension/src/popup/context/routerContext.tsx
@@ -1,4 +1,5 @@
 import React, { createContext, useContext, useState } from 'react'
+import { StorageService } from '@webmonetization/wext/services'
 
 import { ROUTES } from '../constants'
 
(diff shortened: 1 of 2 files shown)


chris-jackson-actionqa (approved):

packages/coil-extension/src/popup/context/routerContext.tsx
```diff
@@ -12,12 +13,33 @@ interface IRouterContext {
...
 const RouterContext = createContext({} as IRouterContext)
 
 // Provider
-export const RouterProvider: React.FC = props => {
-  const [route, setRoute] = useState<string>(ROUTES.streaming)
+export const RouterProvider: React.FC<IRouterProvider> = props => {
+  const { storage } = props
+  const userObject = storage.get('user')
+
+  // Determine the default view based on if the user is allowed to tip
```
sublimator, line 28: Is there anyway we could have this remember your last used view in dev mode ? 
This makes things kind of annoying for testing

PR #2836, opened 2022-05-18
fix(coil-extension): special case coilSite, remove Router key prop

sublimator:


mwill8886 (approved):
Looks great, thanks for fixing these 👍

PR #2864, opened 2022-05-24
Replace global localStorage with props storage

Danisco212:


packages/coil-extension/src/popup/mocks/loadMockedStates.tsx
```diff
@@ -191,6 +197,18 @@ const payingTwitch = mockState({
   adapted: true
 })
 
+const payingFacebook = mockState({
```
sublimator, line 200: add one state where it's showing tipping tab  (this looks it does)
and another where it's showing streaming tab

PR #2865, opened 2022-05-24
Patch deprecated debug.destroy with yarn patch

Danisco212:


packages/coil-extension/package.json
```diff
@@ -67,7 +67,7 @@
     "@webmonetization/polyfill-utils": "^0.0.0",
     "@webmonetization/types": "^0.0.0",
     "@webmonetization/wext": "^0.0.0",
-    "debug": "^4.3.1",
+    "debug": "patch:debug@npm%3A4.3.4#~/.yarn/patches/debug-npm-4.3.4-4513954577.patch",
```
sublimator, line 70: just leave this is as normal, you only need to touch the root package.json

the `resolutions` field does an automagical override

.yarn/patches/debug-npm-4.3.4-4513954577.patch
```diff
@@ -0,0 +1,70 @@
+diff --git a/src/browser.js b/src/browser.js
+index cd0fc35d1ee11e0d6e15421021a54c18958e04d9..ac3f7e1339b9852e42b502eb0c049a32c455c5b0 100644
+--- a/src/browser.js
++++ b/src/browser.js
+@@ -9,16 +9,6 @@ exports.save = save;
+ exports.load = load;
+ exports.useColors = useColors;
+ exports.storage = localstorage();
+-exports.destroy = (() => {
```
sublimator, line 9: We don't want to completely remove it, just make the function a `noop` (no op)eration) 
Because somewhere in our code (or a dependency library) it's being called

.yarn/patches/debug-npm-4.3.4-4513954577.patch
```diff
@@ -0,0 +1,15 @@
+diff --git a/src/browser.js b/src/browser.js
+index cd0fc35d1ee11e0d6e15421021a54c18958e04d9..c0cfdf294fb0428c4d7f4c41ad641c8027e15197 100644
+--- a/src/browser.js
++++ b/src/browser.js
+@@ -11,7 +11,9 @@ exports.useColors = useColors;
+ exports.storage = localstorage();
+ exports.destroy = (() => {
+ 	let warned = false;
+-
++	if(!warned){
```
sublimator, line 10: Did you test this ?

sublimator: image

Looks like it works, though I don't really know how

PR #2884, opened 2022-05-30
Update puppeteer page selector for sign out button in logoutCoil.ts

Danisco212:


packages/coil-puppeteer-utils/src/lib/logoutCoil.ts
```diff
@@ -1,5 +1,6 @@
 import { Page } from 'puppeteer'
 
+import { addCloudFlareAccessHeaders } from './addCloudFlareAccessHeaders'
```
sublimator, line 3: unused import

packages/coil-puppeteer-utils/src/lib/logoutCoil.ts
```diff
@@ -9,11 +10,18 @@ export async function logoutCoil(coilPage: Page): Promise<Page> {
   await coilPage.goto(`${COIL_DOMAIN}/settings`)
   await timeout(2e3)
 
-  const menuSelector = `svg[data-cy='hamburger-toggle']`
-  const logoutSelector = "//p[contains(., 'Log out')]"
+  await addCloudFlareAccessHeaders(coilPage)
```
sublimator, line 13: Oh, it's used here. But note that it's unneeded

packages/coil-puppeteer-utils/src/lib/logoutCoil.ts
```diff
@@ -9,11 +10,18 @@ export async function logoutCoil(coilPage: Page): Promise<Page> {
...
+
+  const menuSelector = `[data-testid='MenuRoundedIcon']`
+  const signButton = "[data-testid='signout-nav-button']"
```
sublimator, line 16: signButton? signOutButton?

sublimator: If you have two stable ids now, what's the need for the xpath?

packages/coil-puppeteer-utils/src/lib/logoutCoil.ts
```diff
@@ -9,11 +10,18 @@ export async function logoutCoil(coilPage: Page): Promise<Page> {
...
 
   try {
     await coilPage.click(menuSelector, { delay: 5e3, clickCount: 1 })
+    const [logoutBtn] = await coilPage.$x(
+      `//div[contains(@class, "MuiListItemButton-root") and contains(., "Sign out")]`
+    )
+    await logoutBtn.click()
+    return coilPage
```
sublimator, line 24: The catch is now catching multiple possible errors, but simply logging "failed to open hamburger menu"

packages/coil-puppeteer-utils/src/lib/logoutCoil.ts
```diff
@@ -9,11 +10,18 @@ export async function logoutCoil(coilPage: Page): Promise<Page> {
...
-  const logoutSelector = "//p[contains(., 'Log out')]"
+  await addCloudFlareAccessHeaders(coilPage)
+
+  const menuSelector = `[data-testid='MenuRoundedIcon']`
+  const signButton = "[data-testid='signout-nav-button']"
 
   try {
     await coilPage.click(menuSelector, { delay: 5e3, clickCount: 1 })
+    const [logoutBtn] = await coilPage.$x(
+      `//div[contains(@class, "MuiListItemButton-root") and contains(., "Sign out")]`
```
sublimator, line 21: You can also try something like this:
`'//*[text()="Sign out"]'`

Ideally we can get the team to put in some stable`data-testid` for ALL the relevant elements 
Failing that, the more resilient it is to change the better. 

Some absurd examples to make the point for them selves:
```
/html/body/div/div/div[1]/div[1]/div/ul[3]/div[2]/div[2]/span
```

```
//*[@id="__next"]/div/div[1]/div[1]/div/ul[3]/div[2]/div[2]/span
```

Both the above are quite "fragile"

PR #2890, opened 2022-05-31
replace ts-node-dev with nodemon

Danisco212:


packages/coil-extension/test.sh
```diff
@@ -36,7 +36,7 @@ function retry() {
 if [[ ${DEV} = 'false' ]]; then
   COMMAND="ts-node -r tsconfig-paths/register -T -P test/tsconfig.json"
 else
-  COMMAND="ts-node-dev -r tsconfig-paths/register -P test/tsconfig.json --respawn --transpile-only"
+  COMMAND="nodemon -r tsconfig-paths/register -P test/tsconfig.json --respawn --transpile-only"
```
sublimator, line 39: did you test this command works ?

sublimator: those options seem particular to `ts-node-x`

sublimator: image

PR #2912, opened 2022-06-03
feat(coil-extension): initial progress on cookie-based auth

kyleaedwards:
TODO:
- [x] image
  - The action icon doesn't update immediately upon logging in
- [x] Check that credentials aren't sent for adapted content queries (youtube/twitch)
  - SuperTokens.init(...) automatically wraps fetch
- [x] #3228
  - Added SuperTokensService which does some hackery to work around this 
- [x] Confirm that the "incognito":"spanning" declaration in the manifest allows the extension to send stored cookies with API requests.
  - [ ] how to handle synchronizing sign ins (and outs) as was done with tokens? 
- [ ] CookieHandlerInterface, it's not just setting one cookie, (99%) must behave like `document.cookie = cookiestring` and merge into document.cookie, rather than replace the property
  - consider extension cookies api with view to maintaining only a single session 
- [ ] window.fetch SuperTokens wrapping is broken image
  - this is in the console, and might not actually be a problem, given references to fetch for `@coil/client` already created
- [x] What happens when cookies are cleared, but coil.com localStorage.token remains? 
  - the extension will delete the token in the content script (running at document_start)
- [ ] Even for MV2, the SuperTokens.init(...) call will at some point trigger a call to use document.cookie (which may not even be a problem for MV2. For MV3, it's possible could use cookies api to get/set cookies on coil.com ) 
  - See: https://github.com/supertokens/supertokens-website/blob/master/lib/ts/fetch.ts#L555
- [ ] browser-tabs-lock requires window/localStorage
- [ ] synchronization of token/cookies ( if you delete cookies at coil.com, the token will keep you logged in)
  - [x] delete token at coil.com in content script
- [ ] #3316 
- [ ] Create custom lock: https://github.com/supertokens/supertokens-website/pull/186


packages/coil-client/src/graphQlClient.ts
```diff
@@ -48,23 +52,57 @@ export class GraphQlClient {
     @inject(GraphQlClientOptions)
     private config: GraphQlClientOptions
   ) {
-    this.fetch = this.config.fetch
+    this.inBrowser = typeof window !== 'undefined'
+    this.fetch = this.inBrowser
```
sublimator, line 56: where had this been a problem?

packages/coil-anonymous-tokens/src/index.ts
```diff
@@ -249,13 +250,13 @@ export class AnonymousTokens implements AnonymousTokensService {
     }
   }
 
-  private async _populateTokensNow(coilAuthToken: string): Promise<void> {
+  private async _populateTokensNow(coilAuthToken?: string): Promise<void> {
```
sublimator, line 253: The changes in `@coil/client` don't seem to support working with either cookies or a token, why is this seemingly different?

packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -120,108 +152,19 @@ export class AuthService extends EventEmitter {
     return this._op
   }
 
-  tokenInvalid(token: string | null) {
-    return !token || tokenUtils.isExpired({ token })
+  initialize() {
+    SuperTokens.init({
```
sublimator, line 156: What is this actually doing ? 

If it's only initiated so the signOut call works, then perhaps it's not even needed? 
Cause the log out will be done on coil.com, no? 
Currently, the extension listens for `coil_writeToken` and when it's an empty string, a logout is inferred. 
It then clears its own copy of the token, so that it doesn't go about injecting it straight back in.
With cookies, there's no real auth state maintained and the store just needs to be cleared of the cached user.

packages/coil-client/src/graphQlClient.ts
```diff
@@ -48,23 +52,57 @@ export class GraphQlClient {
...
+            },
+            {
+              id: 'password',
+              value: password
+            }
+          ]
+        })
+      }
+    )
+    const data = await response.json()
```
sublimator, line 87: there is `response.ok`
generally good to check before looking at the body (.json() )

packages/coil-client/src/queries/queryToken.ts
```diff
@@ -18,6 +18,6 @@ export const queryTokenQuery = `{
   }
 }`
 
-export async function queryToken(this: GraphQlClient, token: string) {
-  return this.query<QueryTokenData>({ query: queryTokenQuery, token })
+export async function queryToken(this: GraphQlClient) {
```
sublimator, line 21: May as well kill this method

packages/coil-extension/src/background/services/AuthService.ts
sublimator: Is this what keeps the cookie fresh when you don't go to coil.com ?

packages/coil-extension/src/background/services/AuthService.ts
```diff
@@ -75,24 +85,46 @@ export class AuthService extends EventEmitter {
...
-  private _op: Promise<string | null> | null = null
+  private _op: Promise<boolean> | null = null
 
   /*
   TODO: manifest version 3 and background workers ?
   If the token is issued more than one day ago, refresh it, such that the
   token is always valid for at least 26-28 days.
    */
-  queueTokenRefreshCheck() {
+  queueAuthRefreshCheck() {
```
sublimator, line 95: Can remove this I think

packages/coil-extension/src/background/services/AuthService.ts
sublimator: https://supertokens.com/docs/website/index.html#sending-requests-with-fetch

So window.fetch is wrapped and monkey patched?

packages/coil-client/src/graphQlClient.ts
kyleaedwards: I'll see if I can try to reproduce the issue, but my browser throws an error if you attempt to call `fetch` unbounded to `window`.

```js
class FetchTest {
    constructor() {
        this.fetch = window.fetch
    }
}
const f = new FetchTest()
f.fetch('/') // Throws "TypeError: Failed to execute 'fetch' on 'Window': Illegal invocation"
```

sublimator: The portableFetch we use everywhere is a wrapper function (wrapper = `(...args) => fetch(...args)` essentially) so this shouldn't be needed. There was a time in main relatively recently where there was a regression circa this though. Maybe you copped that.

packages/coil-extension/src/background/services/AuthService.ts
sublimator: So with that wrapping, we shouldn't even really need to add the `credentials:include` 
What does it actually do when it wraps the function?
SuperTokens supports bearer tokens as well as cookies right? 
So the wrapping would include those ? 

In any case (if we haven't already) we should probably put the SuperTokens import inside background.ts, right at the top, before other imports. 

> we may have already included this in our /signout path

As it stands, the extension doesn't have a dedicated logout UI.  It just tracks the state of coil.com, listening to logout events (coil_writeToken with an empty string)

How will the extension know a logout has happened now? Is there a logout event emitted?

kyleaedwards: You're right about the credentials, looks like unless you explicitly say otherwise, it'll [automatically include them](https://github.com/supertokens/supertokens-website/blob/10285c14dd8a47b72098175f7f473adf3ffaf8cc/lib/ts/utils/index.ts#L99-L102).

With SuperTokens, it's a little funky but the access token cookie by default is a JWT-like token (but it's missing a few key props [unless modified](https://supertokens.com/docs/thirdpartyemailpassword/common-customizations/sessions/with-jwt/about)). It is possible to [store this token in localStorage](https://supertokens.com/docs/thirdpartyemailpassword/advanced-customizations/examples/localstorage/about#frontend), but it looks like you have to modify the frontend manually.

> How will the extension know a logout has happened now? Is there a logout event emitted?

If you were to log out from another window, it wouldn't be immediately apparent to the extension until the next time it attempted an API request (reading whoami data, fetching an anonymous token, tipping, etc...) in which case the browser would not have any access/refresh token cookies to pass along and the request would fail.

packages/coil-client/src/graphQlClient.ts
kyleaedwards: I'll pull that out and see if it's still happening, I bet you're right that I just caught that regression.

packages/coil-anonymous-tokens/src/index.ts
sublimator: There's a bunch of errors as @coil/client is used outside the extension.

On that note, if you wanted to, you could just make token optional for the moment in all the client helper methods, until we decide to update all calling sites (at some later point - adding a cookie impl to the client ). In terms of calling sites, there's a command line script that connects via user/pass and streams to a payment pointer specified on the command line for example. Probably a few others.

packages/coil-extension/src/background/services/AuthService.ts
sublimator: > If you were to log out from another window, it wouldn't be immediately apparent to the extension
Can we add an event then?
At present, when we logout, we update the browser action icon immediately

kyleaedwards: Does it do that by constantly polling localStorage? And if so, does it decrypt the JWT each time to know when the token is expired as well?

sublimator: No, we bind to the `coil_writeToken` event from in the content script ( extension component that runs within a page ) when it's running under coil.com. 

We check once on startup to catch logouts that happened while the extension was temporarily disabled.

kyleaedwards: I just checked and we're still dispatching that event on the signout page (I'm putting in a change to make sure that occurs AFTER we perform the SuperTokens signout).

I can't change the name of the event because of the current versions of the extension. Likewise, we should be sending the event on signin. It looks like it is but I'll double-check with Mike tomorrow to make sure.

Right now we're still setting the legacy JWT so this should all still work, but in the long run, I guess we should update the content script's `syncViaInjectToken`/`injectToken` communication with the background script to do something like check the user via the API, or use some non-token localStorage flag whether the user is authed.

sublimator: Would be good to have a comment on what the init call does (wraps fetch, refreshes tokens automatically etc)

sublimator:
@kyleaedwards 
Yah, it was writing to the wrong store and clearing from the correct one!

packages/coil-client/src/queries/tip.ts
```diff
@@ -25,12 +25,19 @@ export interface TipInput {
 
 export async function tip(
   this: GraphQlClient,
-  token: string,
-  input: TipInput
+  token: string | TipInput,
+  input?: TipInput
```
sublimator, line 29: It is probably better here to just change the function signature to put input first

sublimator: The only "clients" for this lib are really in this repo. The package isn't published on npm or anything, so it's fine to break back-compat.

PR #3101, opened 2022-07-06
refactor(coil-extension): bind all events at startup

sublimator:


packages/coil-extension/src/background/services/BackgroundEvents.ts
```diff
@@ -0,0 +1,130 @@
...
+  'tabs.onRemoved': chrome.tabs.onRemoved,
+  'tabs.onReplaced': chrome.tabs.onReplaced,
+  'tabs.onActivated': chrome.tabs.onActivated,
+  'tabs.onActiveChanged': chrome.tabs.onActiveChanged,
+  'tabs.onSelectionChanged': chrome.tabs.onSelectionChanged,
+  'tabs.onCreated': chrome.tabs.onCreated,
+  'tabs.onAttached': chrome.tabs.onAttached,
+  'tabs.onDetached': chrome.tabs.onDetached,
+
+  // TODO: this will need special handling ...
```
sublimator, line 17: // Don't need to return true here, not using sendResponse
        // https://developer.chrome.com/apps/runtime#event-onMessage

PR #3114, opened 2022-07-08
mv3 rush ahead

sublimator:


packages/coil-extension/webpack.common.ts
```diff
@@ -9,6 +11,9 @@ const data = Buffer.from(wmPolyfill, 'utf-8')
 const digest = createHash('sha256').update(data).digest()
 const polyfillHash = `sha256-${digest.toString('base64')}`
 
+// write polyfill
+fs.writeFileSync(`${__dirname}/dist/${polyfillHash}.js`, data)
```
sublimator, line 15: Use a webpack method

packages/coil-extension/src/background/di/configureContainer.ts
```diff
@@ -38,7 +38,7 @@ export function configureContainer({
   }
 
   container.bind(tokens.CoilDomain).toConstantValue(coilDomain)
-  container.bind(tokens.UserAgent).toConstantValue(navigator.userAgent)
+  container.bind(tokens.UserAgent).toConstantValue('navigator.userAgent')
```
sublimator, line 41: WorkerNavigator should be available ??

sublimator: So the problem here was that `Navigator` when used as an injection symbol was no longer available in the worker, as that class is not defined.

packages/coil-extension/src/background/services/utils.ts
```diff
@@ -53,7 +53,7 @@ export function createLogger(context: Context) {
   }
   const namespace = `background${name ? `:${name}` : ''}`
   // eslint-disable-next-line no-console
-  if (localStorage.debug) {
+  if (self.localStorage?.debug) {
```
sublimator, line 56: Check if debug has been updated and has some config mechanism for workers

PR #3128, opened 2022-07-13
fix(webpack-utils): make it clear whether .git is present or not

sublimator:


packages/coil-webpack-utils/src/index.ts
```diff
@@ -34,31 +34,46 @@ export const configureNodePolyfills = (wpConf: webpack.Configuration) => {
...
-    version,
-    commit: hash.toString().trim(),
-    dirty: status.toString().trim().length > 0,
+  const withoutGit = {
+    haveGitFolder: false,
+    version: parsed.version as string,
     buildDateISO: date.toISOString()
+  } as const
+
+  if (fs.existsSync(gitPath)) {
```
sublimator, line 65: this won't actually work, because .git is in the root of the folder ...

PR #3129, opened 2022-07-13
chore: github action to save extension build and source code

Danisco212:


.github/workflows/archive-extension-source-code.yml
```diff
@@ -0,0 +1,77 @@
...
+    steps:
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: source-${{env.RELEASE_VERSION}}
+
+      - name: Step 1 - Extract tar archive
+        run: tar -xf monorepo.tar --exclude=**/.git/*
```
sublimator, line 40: the --exclude option should be on the creation of the tar

.github/workflows/archive-extension-source-code.yml
```diff
@@ -0,0 +1,77 @@
...
+  save-monorepo-source-code:
+    runs-on: ubuntu-latest
+   
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
+      - name: Step 1 - Tar files
```
sublimator, line 18: Add a comment with a link to the permissions issues

.github/workflows/archive-extension-source-code.yml
```diff
@@ -0,0 +1,77 @@
...
+        run: tar --exclude='.git' -cf monorepo.tar ${{ github.workspace }}
```
sublimator, line 19: Add a `-C`
tar --exclude='.git' -cf monorepo.tar -C `${{ github.workspace }}`

I think ... try it locally

sublimator: You need a `.` at the end

.github/workflows/archive-extension-source-code.yml
sublimator: Zip format would actually work fine (if you used a zip util), and would be preferable (you could download the zip inside a zip, extract the zip and upload it. with tar, the reviewer will have to extract the tar as an extra step) 

It's the upload/download actions that are the problem here.

PR #3285, opened 2022-09-08
feat(coil-extension): initial progress on cookie-based auth

sublimator:


packages/coil-client/src/graphQlClient.ts
```diff
@@ -48,35 +54,88 @@ export class GraphQlClient {
     @inject(GraphQlClientOptions)
     private config: GraphQlClientOptions
   ) {
+    this.inBrowser = typeof window !== 'undefined'
```
sublimator, line 57: TODO: mv3

PR #3344, opened 2022-09-18
fix: acount corrected to account

johnnyreilly:
This PR corrects the typo in the extension that routes to `acount` instead of `account`.  It appears that it's already fixed on the backend as you can see that when you arrive at `acount` you're redirected to `account`.  This PR should make that redirection unnecessary. The GIF below illustrates what happens now, should this PR be merged, the redirect you see at present should not happen.

acount-to-account


sublimator:
@johnnyreilly 

Thanks!

PR #3672, opened 2023-02-27
feat: disable everything, show only monetization status

sublimator:


sublimator:
Ok, I  had thought you just wanted this in a branch

traviscrist:
@sublimator that was my initial idea but I'd rather we merge this in since then main is in the final state that we have an up to date version of the extension of which is released.

sublimator:
Should be able to get to it today or tomorrow (my time)

PR #3754, opened 2023-04-20
feat: move to pnpm

sublimator:
Yarn 2 pnp doesn't fully support ESM when using extra loaders like ts-node/esm.
Up until a recent version of Node 19 there was no support for the chaining of
loaders required to support both pnp and on-the-fly TypeScript compilation at
once.

The @dummy/ext is the first ESM enabled package in the mono repo, but it was 
found that the polyfills were not working correctly when using pnp. To solve
this we have two options:

1. Just move to pnp with nodeLinker: node-modules
2. Move to pnpm

Given the work has already been done (working branch with a port to PNPM) and
that ILF uses PNPM already in the Rafiki/InterledgerJS repos, it probably makes
sense to go ahead and use PNPM.

One big downside is the substantial performance degradation moving away from
pnp, which despite its at times brittle nature, really is far superior in
performance.


packages/webexts-build-utils/scripts/build.sh
```diff
@@ -54,5 +54,5 @@ else
 fi
 
 echo Running ${BUILD_ENV} build for ${BROWSER_NAME}...
-# See yarn plugin for TSCONFIG_PATHS_REGISTER
-TS_NODE_TRANSPILE_ONLY=${TS_NODE_TRANSPILE_ONLY:-true} BROWSER=${BROWSER_NAME} API=${BROWSER_API} yarn run webpack-cli --config ${WEBPACK} ${EXTRA_ARGS}
+# See pnpm plugin for TSCONFIG_PATHS_REGISTER
```
sublimator, line 57: TODO

.github/workflows/generated-jobs.yml
```diff
@@ -22,21 +22,26 @@ jobs:
...
+        uses: pnpm/action-setup@v2
+        with:
+          version: 8
       - name: Restore Cache
         uses: actions/cache@v3
         with:
-          path: ${{ github.workspace }}/.yarn
-          key: v1-dependencies-${{ runner.os }}-${{ hashFiles('package.json', 'yarn.lock') }}-${{ matrix.node-version }}
-      - name: Yarn Install
+          path: ${{ github.workspace }}/node_modules
```
sublimator, line 32: TODO, what about 
- /home/circleci/.local/share/pnpm/store/v3
-puppeteer-cache

PR #3911, opened 2023-06-27
feat: jest with esm support

sublimator:
## Changes

1. **Unified `jest.config.cjs`**: We've simplified our Jest setup by deploying one `jest.config.cjs` file in each workspace. This file now treats `.mts` files as ESM by adding them to `extensionsToTreatAsEsm` and `moduleFileExtensions`. The configuration now also collects all workspace Jest configurations as `projects`.

2. **Removal of `jest-e2e.config.cjs`**: We've deleted the `jest-e2e.config.cjs` files, further streamlining the configuration process.

3. **Usage of `PROJECT_JEST` Variable**: The `PROJECT_JEST` environment variable should be set in the package.json scripts in each workspace. This signals to Jest when workspace tests are run directly, rather than being collected from the monorepo root.

4. **Update in `package.json` Test Scripts**: In the test scripts of the root `package.json`, we've set the `NODE_OPTIONS` variable to `--experimental-vm-modules`.

5. **Removal of `@types/jest`**: We've removed `@types/jest` and now import globals from `@jest/globals` instead.

By employing this strategy, we've streamlined the Jest configuration process for ESM support, improved the manageability of Jest setup for each workspace, and facilitated a smoother, more gradual transition from CJS to ESM.


jest.config.mjs
```diff
@@ -0,0 +1,56 @@
...
+        // Speed up typechecks
+        skipLibCheck: SKIP_LIB_CHECK
+      },
+      isolatedModules: ISOLATED_MODULES,
+      diagnostics: false
+    }
+  }
+}
+
+const localConfPath = `${ROOT}/jest.config.local.cjs`
```
sublimator, line 50: TODO

